PR #6, opened 2023-01-26
[Lee Tze Han] iP

ltzehan:


src/main/java/Duke.java
```diff
@@ -1,10 +1,136 @@
...
+    /**
+     * User commands
+     */
+    private static final String COMMAND_EXIT = "bye";
+    private static final String COMMAND_LIST = "list";
+    private static final String COMMAND_MARK = "mark";
+    private static final String COMMAND_UNMARK = "unmark";
+    private static final Set<String> COMMANDS_TASK = new HashSet<>(Arrays.asList("todo", "deadline", "event"));
+
+    private static List<Task> userTasks = new ArrayList<>();
```
mitchmalvin1: Is there a need to use ArrayList?

src/main/java/Duke.java
```diff
@@ -1,10 +1,136 @@
...
+
+        } else if (cmd.equals("event")) {
+            // Syntax: event <desc> /from <start> /to <end>
+            String[] params = splitInput[1].split("(/from|/to)");
+            task = new EventTask(params[0].trim(), params[1].trim(), params[2].trim());
+
+        }
+
+        userTasks.add(task);
+        System.out.println("Got it. I've added this task:");
```
mitchmalvin1: Should this chunk of print functions be abstracted to another function?

src/main/java/Duke.java
```diff
@@ -1,10 +1,136 @@
...
+
     public static void main(String[] args) {
-        String logo = " ____        _        \n"
-                + "|  _ \\ _   _| | _____ \n"
-                + "| | | | | | | |/ / _ \\\n"
-                + "| |_| | |_| |   <  __/\n"
-                + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
+        printDivider();
+        System.out.println("Hello! I'm Duke\nWhat can I do for you?");
```
mitchmalvin1: Can this be moved to a new greeting() function?

src/main/java/Duke.java
```diff
@@ -1,10 +1,136 @@
...
+        for (int index = 1; index <= userTasks.size(); index++) {
+            System.out.println(index + ". " + userTasks.get(index - 1));
+        }
+    }
+
+    private static void setUserTaskState(int userIndex, boolean isDone) {
+        // Tasks are 0-indexed, user index is 1-indexed
+        int index = userIndex - 1;
+        if (index < 0 || index >= userTasks.size()) {
+            System.out.println("Oops, not quite sure what task you're referring to...");
```
mitchmalvin1: Shouldn't this throw an Exception?

src/main/java/Duke.java
```diff
@@ -1,10 +1,136 @@
...
 public class Duke {
+
+    /**
+     * User commands
+     */
+    private static final String COMMAND_EXIT = "bye";
+    private static final String COMMAND_LIST = "list";
+    private static final String COMMAND_MARK = "mark";
+    private static final String COMMAND_UNMARK = "unmark";
+    private static final Set<String> COMMANDS_TASK = new HashSet<>(Arrays.asList("todo", "deadline", "event"));
```
mitchmalvin1: This is good considering how there can be more types of task in the future.

src/main/java/Duke.java
```diff
@@ -1,10 +1,136 @@
...
+            task = new Task(splitInput[1]);
+
+        } else if (cmd.equals("deadline")) {
+            // Syntax: deadline <desc> /by <deadline>
+            String[] params = splitInput[1].split("/by");
+            task = new DeadlineTask(params[0].trim(), params[1].trim());
+
+        } else if (cmd.equals("event")) {
+            // Syntax: event <desc> /from <start> /to <end>
+            String[] params = splitInput[1].split("(/from|/to)");
```
mitchmalvin1: Should you put the magic string "(/from|/to)" into a final constant?

src/main/java/Duke.java
```diff
@@ -1,16 +1,23 @@
...
 public class Duke {
 
     /**
      * User commands
      */
-    private static final String COMMAND_EXIT = "BYE";
-    private static final String COMMAND_LIST = "LIST";
-    private static final String COMMAND_MARK = "MARK";
-    private static final String COMMAND_UNMARK = "UNMARK";
+    private static final String COMMAND_EXIT = "bye";
```
sistine-yu: it's a nice try to name the command string variable

src/main/java/Duke.java
```diff
@@ -50,14 +57,40 @@ private static void setUserTaskState(int userIndex, boolean isDone) {
         }
     }
 
+    private static void handleAddUserTask(String cmd, String[] splitInput) {
+        Task task = null;
```
sistine-yu: maybe there's no need to create a null Task object first

src/main/java/Duke.java
```diff
@@ -73,11 +106,12 @@ private static void handleUserInput(String userInput) {
             int index = Integer.parseInt(splitInput[1]);
             setUserTaskState(index, false);
 
+        } else if (COMMANDS_TASK.contains(cmd)) {
+            handleAddUserTask(cmd, splitInput);
+
         } else {
-            Task task = new Task(userInput);
-            userTasks.add(task);
+            System.out.println("Sorry, I don't recognize that command...");
```
sistine-yu: it's nice to handle the unrecognized cases

src/main/java/task/Task.java
```diff
@@ -14,6 +16,6 @@ public void setTaskState(boolean isDone) {
 
     @Override
     public String toString() {
-        return "[" + (isDone ? "x" : " ") + "] " + description;
+        return "[T][" + (isDone ? "x" : " ") + "] " + description;
```
sistine-yu, line 45: maybe try to separate the TodoTask as well to make it more organized

src/main/java/Duke.java
```diff
@@ -1,10 +1,136 @@
...
+
+        } else if (cmd.equals(COMMAND_UNMARK)) {
+            int index = Integer.parseInt(splitInput[1]);
+            setUserTaskState(index, false);
+
+        } else if (COMMANDS_TASK.contains(cmd)) {
+            handleAddUserTask(cmd, splitInput);
+
+        } else {
+            System.out.println("Sorry, I don't recognize that command...");
```
WilsonLee2000: I like how you have implemented the "if", "else if" and "else" in such a way that it maximises readability and makes your code easier for audiences to read! :)

src/main/java/Duke.java
```diff
@@ -1,10 +1,136 @@
...
+    private static void handleAddUserTask(String cmd, String[] splitInput) {
+        Task task = null;
+
+        if (cmd.equals("todo")) {
+            // Syntax: todo <desc>
+            task = new Task(splitInput[1]);
+
+        } else if (cmd.equals("deadline")) {
+            // Syntax: deadline <desc> /by <deadline>
+            String[] params = splitInput[1].split("/by");
```
WilsonLee2000: I like how you named multi-valued variables using a plural term as it allows readers to distinguish clearly between single-valued and multi-valued variables.

src/main/java/task/EventTask.java
```diff
@@ -0,0 +1,19 @@
+package task;
+
+public class EventTask extends Task {
+
+    protected String start;
+    protected String end;
+
+    public EventTask(String description, String start, String end) {
+        super(description);
+
```
WilsonLee2000, line 23: Perhaps there is not a need for the space (line 10) to be here?

src/main/java/Duke.java
```diff
@@ -1,10 +1,136 @@
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Scanner;
+import java.util.Set;
+import task.DeadlineTask;
+import task.EventTask;
+import task.Task;
```
WilsonLee2000: I really like how you only import the relevant libraries that are needed! Instead of using * import!

WilsonLee2000 (commented):
Overall, well done on your code quality! I'm really impressed! Just a minor additional spacing issue, kindly refer to one of my comments. :)

src/main/java/storage/TaskStorageManager.java
```diff
@@ -0,0 +1,70 @@
...
+                Files.createFile(filePath);
+            } catch (IOException ex) {
+                System.out.println("Couldn't save your files to: " + filePath + "!");
+                ex.printStackTrace();
+                return;
+            }
+        }
+
+        // Serialize and write to file
+        try (ObjectOutputStream outputStream = new ObjectOutputStream(new FileOutputStream(filePath.toString()))) {
```
wamikamalik: You can try splitting this into multiple lines of code inside the try statement to improve readability. Similar issue in line 61.

src/main/java/task/DeadlineTask.java
```diff
@@ -0,0 +1,32 @@
...
+
+    public DeadlineTask(String description, String deadline) {
+        super(description);
+
+        this.deadline = deadline;
+    }
+
+    public static DeadlineTask createFromInput(String[] splitInput) throws InvalidSyntaxException {
+        try {
+            String[] params = splitInput[1].split("/by");
```
wamikamalik: Maybe refactor this magic literal to a constant for better readability. For more info refer to [Code Quality - Magic Numbers](https://nus-cs2113-ay2122s1.github.io/website/se-book-adapted/chapters/codeQuality.html#avoid-magic-numbers) on the course website. Similar changes can be made in other places with magic literals.

src/main/java/Duke.java
```diff
@@ -1,10 +1,150 @@
...
+            if (cmd.equals(Command.MARK.label)) {
+                setUserTaskState(index, true);
+            } else if (cmd.equals(Command.UNMARK.label)) {
+                setUserTaskState(index, false);
+            } else if (cmd.equals(Command.DELETE.label)) {
+                Task removedTask = userTasks.remove(index);
+
+                System.out.println("Alrighty, I've removed this task:");
+                System.out.println(removedTask);
+                printUserTasksCount();
```
wamikamalik: Maybe refactor the first two print statements as well. Try to be consistent with the level of abstraction, i.e., try to keep the same level of abstraction within a code fragment.

src/main/java/Duke.java
```diff
@@ -1,10 +1,150 @@
...
+            int userIndex = Integer.parseInt(splitInput[1]);
+            int index = userIndex - 1;
+
+            if (cmd.equals(Command.MARK.label)) {
+                setUserTaskState(index, true);
+            } else if (cmd.equals(Command.UNMARK.label)) {
+                setUserTaskState(index, false);
+            } else if (cmd.equals(Command.DELETE.label)) {
+                Task removedTask = userTasks.remove(index);
+
```
wamikamalik: Maybe remove this whitespace as it could be a little misleading.

src/main/java/Duke.java
```diff
@@ -1,10 +1,150 @@
...
+import storage.TaskStorageManager;
+import task.DeadlineTask;
+import task.EventTask;
+import task.Task;
+import ui.Command;
+
 public class Duke {
+
+    private static final TaskStorageManager taskStorageManager = new TaskStorageManager();
+    private static final List<Task> userTasks = taskStorageManager.loadTasks();
```
wamikamalik: Maybe try changing this bit. We use final for constants, so those variables should be named in all caps according to our Java coding standards. If these variables are not final, u should remove the keyword final. For example, I see that you are adding tasks to userTasks, so this list cannot be final.

src/main/java/exceptions/InvalidSyntaxException.java
```diff
@@ -0,0 +1,15 @@
+package exceptions;
+
+public class InvalidSyntaxException extends Exception {
+
+    private final String expectedSyntax;
```
wamikamalik, line 10: If this string is not a constant, u should remove the "final". If it is indeed a constant, the name should be all uppercase using underscore to separate words. Similar issue in other places.

src/main/java/storage/TaskStorageManager.java
```diff
@@ -0,0 +1,70 @@
...
+    public void saveTasks(List<Task> tasks) {
+        // Check for file existence
+        if (!Files.isRegularFile(filePath)) {
+            try {
+                Files.createDirectories(filePath.getParent());
+                Files.createFile(filePath);
+            } catch (IOException ex) {
+                System.out.println("Couldn't save your files to: " + filePath + "!");
+                ex.printStackTrace();
+                return;
```
wamikamalik: Is this return needed here?

src/main/java/task/DeadlineTask.java
```diff
@@ -0,0 +1,32 @@
+package task;
+
+import exceptions.InvalidSyntaxException;
+import ui.Command;
+
+public class DeadlineTask extends Task {
+
+    // For serialization
+    private static final long serialVersionUID = (1 << 5) + 1;
```
wamikamalik, line 12: Remember to name your constants in all uppercase using underscore to separate words. Check for similar issue in other parts of the code.

src/main/java/ui/Command.java
```diff
@@ -0,0 +1,39 @@
...
+    MARK("mark", "mark <task #>"),
+    UNMARK("unmark", "unmark <task #>"),
+    DELETE("delete", "delete <task #>"),
+    TODO("todo", "todo <description>"),
+    DEADLINE("deadline", "deadline <description> /by <deadline>"),
+    EVENT("event", "event <description> /from <start> /to <end>");
+
+    // All commands that create a new task
+    public static final Set<String> ADD_TASK_COMMANDS = new HashSet<>(
+        Arrays.asList(TODO.label, DEADLINE.label, EVENT.label));
```
wamikamalik: Follow the Java coding standard here, "Indentation for wrapped lines should be 8 spaces". Similar issue  below.

wamikamalik (commented):
Great Job! Just a few nits here and there. Do take a look!

src/main/java/storage/TaskStorageManager.java
ltzehan: Thanks for the comment! This is an early-return pattern; if the file/parent directories cannot be created, we do not expect the subsequent saving to the file to be successful, and so this return is required to terminate `saveTasks` early

src/main/java/task/DeadlineTask.java
ltzehan: Thanks for pointing this out! Unfortunately, the field name is mandated by Java's `Serializable` interface which has been implemented by `Task` [(docs)](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/io/Serializable.html)

src/main/java/Duke.java
ltzehan: This is a good point; thanks for raising this! 

Since `taskStorageManager` and `userTasks` are not immutable objects, I do not think these variables should be stylized as constants. I'm not sure if this is explicitly stated in CS2113's guidelines, but an example of this would be [Google's Java Style Guide](https://google.github.io/styleguide/javaguide.html#s5.2.4-constant-names).

Nonetheless, this was very helpful since I realized some of my constants in other areas were in fact mutable!

src/main/java/exceptions/InvalidSyntaxException.java
ltzehan: Thanks for the comment, but I think this is alright -- `expectedSyntax` here is a blank final variable which can, and must, only be initialized in the constructor of `InvalidSyntaxException`. I think together with the private access modifier, this makes it clear that each instance of `InvalidSyntaxException` is associated with a specific `expectedSyntax`, and that it must not be modified after the constructor

PR #9, opened 2023-01-26
[Benjamin Toh] iP

bentohset:


src/main/java/Deadline.java
```diff
@@ -0,0 +1,13 @@
+public class Deadline extends Task{
+    protected String by;
```
nichyjt: There are some inconsistencies between the access modifiers between the subclasses for the Task feature. Some are set as private (e.g. Event) while others are set as protected (s.a. this file). I am unsure if there is a particular intent for this, but if there is none perhaps it would be better practice to standardise the access modifiers across subclasses.

src/main/java/Task.java
```diff
@@ -0,0 +1,25 @@
+public class Task {
+    protected String description;
+    protected boolean isDone;
+
+    public Task(String description){
+        this.description = description;
+        this.isDone = false;
+    }
+
+    public String getStatusIcon(){
```
nichyjt: For the subclasses for Task, there are minor space formatting issues for curly braces. If they come after parentheses "()", there should be a space between it and the opening curly brace.

[Reference for this format](https://se-education.org/guides/conventions/java/basic.html#layout)!

src/main/java/Task.java
```diff
@@ -0,0 +1,25 @@
...
+        return (isDone ? "X" : " ");    //mark task with X or tick
+    }
+
+    public void markDone(){
+        this.isDone = true;     //mark task as done
```
nichyjt: I like how there are comments given thoughout the program. I think to improve on this good practice one minor change can be made.

It might be better to leave [comments in their own lines](https://se-education.org/guides/conventions/java/basic.html#comments) with the same indentation level. 
Although the comments are short, giving them their own lines may be better practice to improve readability in the long run!

nichyjt (commented):
Overall, the code is well factored and follows the SLAP principle well. Apart from minor nits with respect to formatting, your code adheres to good Coding Standards. Good job!

src/main/java/Manager.java
```diff
@@ -0,0 +1,155 @@
...
+
+        case "deadline":
+            if (!inputArray[1].contains("/by")) {
+                System.out.println("\tWrong format! Please include /by.");
+                break;
+            }
+
+            String[] deadlineDetails = inputArray[1].split(" /by ", 2);
+            String deadlineDescription = deadlineDetails[0];
+            String by = deadlineDetails[1];
```
GohJW: Perhaps these lines of code can be implemented in the addDeadline() method to improve readability. Similarly for the case "event" below.

src/main/java/Manager.java
```diff
@@ -0,0 +1,155 @@
...
+            addEvent(eventDescription, from, to);
+            break;
+
+        case "list":
+            printTasks();
+            break;
+
+        case "mark":
+            int taskNum = Integer.parseInt(inputArray[1]);
+            markTask(taskNum);
```
GohJW: Good job extracting methods for easier readability

src/main/java/Duke.java
```diff
@@ -1,10 +1,8 @@
...
     public static void main(String[] args) {
-        String logo = " ____        _        \n"
-                + "|  _ \\ _   _| | _____ \n"
-                + "| | | | | | | |/ / _ \\\n"
-                + "| |_| | |_| |   <  __/\n"
-                + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
+
+        Io.printWelcome();
+        Io.getInput();
```
GohJW: Good use of an IO class to differentiate methods for inputs and printing.

src/main/java/Manager.java
```diff
@@ -0,0 +1,155 @@
...
+        Task deadline = new Deadline(inputMessage, by);
+        if (listSize >= 100) {
+            System.out.println("List size exceeded!");
+            return;
+        }
+        taskList[listSize] = deadline;
+        listSize++;
+
+        String addedMessage = "\tGot it. I've added this task:\n"
+                + "\t  " + deadline.toString();
```
GohJW: Good job using line breaks in order to prevent long lines of code.

src/main/java/Deadline.java
bentohset: Noted on your observation! I agree that such access modifiers should be consistent throughout.

src/main/java/Task.java
bentohset: Thanks for your keen observation, a hindsight on my part!

src/main/java/Task.java
bentohset: Yes i totally agree with your comment. Left it there for debugging purposes and completely forgot about it!

PR #22, opened 2023-01-27
[Sherlock-YH] iP

Sherlock-YH:


src/main/java/Event.java
```diff
@@ -0,0 +1,14 @@
+public class Event extends Task{
+
+    protected String from;
+    protected String to;
+
+    public Event(String description, int num,String from, String to){
```
EthanYidong:
```suggestion
    public Event(String description, int num, String from, String to) {
```
It may improve readability to standardize spacing according to [Google's Coding Standards](https://google.github.io/styleguide/javaguide.html#s4.6.2-horizontal-whitespace). I noticed the same issue in other places too.

src/main/java/Duke.java
```diff
@@ -1,10 +1,171 @@
+import java.util.Scanner;
+
 public class Duke {
+
+    public static final String LOGO = " ____        _        \n"
```
EthanYidong: I like the usage of capitalization for constants.

src/main/java/Task.java
```diff
@@ -0,0 +1,20 @@
+public class Task {
+    protected String description;
+    protected int num;
+    protected boolean isDone;
```
EthanYidong: I like the naming of your boolean variables indicating that they are booleans using "is".

src/main/java/Duke.java
```diff
@@ -1,10 +1,171 @@
...
+            } else if (userCommand.contains("todo")) {
+                addTodoTask(list, counter, userCommand);
+                counter++;
+            } else if (userCommand.contains("deadline")) {//Task type D
+                addDeadlineTask(list, counter, userCommand);
+                counter++;
+            } else if (userCommand.contains("event")) {
+                addEventTask(list, counter, userCommand);
+                counter++;
+            }else{
```
EthanYidong:
```suggestion
            } else {
```
Perhaps add spacing around your else for better readability?

EthanYidong (commented):
Looks good overall, however some coding standards need to be more consistently applied. Specifically, consider inserting spaces between certain symbols to improve readability.

src/main/java/Duke.java
```diff
@@ -1,10 +1,171 @@
...
+            } else if (userCommand.contains("unmark")) {
+                unmarkTask(list, userCommand, counter);
+            } else if (userCommand.contains("mark")) {
+                markTask(list, userCommand,counter);
+            } else if (userCommand.contains("todo")) {
+                addTodoTask(list, counter, userCommand);
+                counter++;
+            } else if (userCommand.contains("deadline")) {//Task type D
+                addDeadlineTask(list, counter, userCommand);
+                counter++;
```
yixuann02: Perhaps you could consider using switch and case for these if-else statements.

src/main/java/Duke.java
```diff
@@ -1,10 +1,171 @@
...
+                System.out.println("Please specify the task you want to mark :) ");
+            }else {
+                int taskNum = Integer.parseInt(subStr);
+                boolean isNumWithinCounter = (taskNum < counter) && (taskNum > 0);
+                if (isNumWithinCounter) {
+                    System.out.println("Okay, I have unmarked Task " + taskNum);
+                    list[taskNum - 1].isDone = false;
+                }else {
+                    System.out.println("Please enter a valid task No. :) ");
+                }
```
yixuann02: Try to avoid deep nesting by splitting it into different methods

src/main/java/Duke.java
```diff
@@ -1,10 +1,171 @@
...
+            } else if (userCommand.contains("event")) {
+                addEventTask(list, counter, userCommand);
+                counter++;
+            }else{
+                invalidEntry();
+            }
+        }
+    }
+
+    private static void invalidEntry() {
```
yixuann02: Good job using multiple methods to make the main readable

src/main/java/Duke.java
```diff
@@ -1,10 +1,171 @@
...
+        String[] arrOfStr = ins.split("deadline");
+        arrOfStr = arrOfStr[1].split("/by ", 2);
+        String name = arrOfStr[0];
+        String by = arrOfStr[1];
+        list[counter - 1] = new Deadline(name, counter, by);
+        System.out.println("Added: " + ins);
+        System.out.println(HORIZONTAL_LINE + "\n");
+    }
+
+    private static void addTodoTask(Task[] list, int counter, String ins) {
```
yixuann02: You did a good job with keeping most of the methods short and readable

src/main/java/commands/AddDeadline.java
```diff
@@ -0,0 +1,36 @@
...
+public class AddDeadline {
+
+    public static void addDeadlineTask(ArrayList<Task> list, String ins) {
+        try {
+
+            if (!ins.contains("/by")) {
+                throw new InvalidTaskException();
+            }
+
+            String[] arrOfStr = ins.split("deadline");
```
kerct, line 23: Instead of arrOfStr which doesn't say anything about what is this variable for, perhaps you can rename it to something like insParts

src/main/java/commands/CreateFile.java
```diff
@@ -0,0 +1,16 @@
...
+import java.io.IOException;
+
+public class CreateFile {
+    public static void createFile(File f) {
+        try {
+            if(f.createNewFile()) {
+                System.out.println("File Created");
+            }
+        } catch (IOException e) {
+            System.out.println("Error occurs");
```
kerct, line 22: Perhaps you can print out what kind of error it is instead of using a generic "error occurs"

src/main/java/commands/Exit.java
```diff
@@ -0,0 +1,24 @@
+package commands;
+
+import tasks.Task;
+
+import java.util.ArrayList;
+import java.io.FileWriter;
+import java.io.IOException;
+
+
```
kerct, line 9:
```suggestion
```

src/main/java/commands/Exit.java
```diff
@@ -0,0 +1,24 @@
...
+                Task task = list.get(i);
+                int isDone = task.getIsDone() ? 1 : 0;
+                fw.write(isDone + " " + task.getUserInput() + System.lineSeparator());
+            }
+            fw.close();
+        } catch (IOException e) {
+            System.out.println("Cannot write to file. Something went wrong");
+        }
+    }
+}
```
kerct: Perhaps you can put this together with Bye, since Exit is not an explicit command

src/main/java/commands/FileIO.java
```diff
@@ -0,0 +1,24 @@
+package commands;
```
kerct, line 1: You might want to consider separating the actual commands like AddDeadline, AddEvent etc. from these files which are not actual commands, by putting them in another package

src/main/java/commands/Greeting.java
```diff
@@ -0,0 +1,13 @@
+package commands;
+
+import constants.constant;
+
+public class Greeting {
+    public static void greeting() {
```
kerct, line 6: When naming a function, verbs should be used. In this case, I feel that this greeting can be made as a constant variable instead of putting it in a function in a separate class on its own.

src/main/java/commands/ListOfTask.java
```diff
@@ -0,0 +1,16 @@
+package commands;
+
+import constants.constant;
+import tasks.Task;
+
+import java.util.ArrayList;
+
+public class ListOfTask {
```
kerct, line 8:
```suggestion
public class ListOfTasks {
```

src/main/java/exceptions/InvalidTaskException.java
```diff
@@ -0,0 +1,7 @@
+package exceptions;
+
+public class InvalidTaskException extends Exception{
```
kerct, line 3:
```suggestion
public class InvalidTaskException extends Exception {
```

kerct (commented):
Code is pretty neat overall, but perhaps you can consider whether it is always necessary to create a file (class) or if it is sufficient to create a variable/ function.

PR #26, opened 2023-01-27
[Poobalan Aatmika Lakshmi] iP

PoobalanAatmikaLakshmi:


src/main/java/Deadline.java
```diff
@@ -0,0 +1,27 @@
...
+    }
+
+    public void setBy(String by) {
+        this.by = by;
+    }
+
+    Deadline(String description, String by) {
+        super(description);
+        setBy(by);
+
```
natashatanyt: Please be consistent with your spacing! ie. most of your code has the closing brace on the immediate next line, but this one has an additional space

src/main/java/Todo.java
```diff
@@ -0,0 +1,12 @@
...
+
+    public Todo(String description) {
+        super(description);
+    }
+
+    @Override
+    public char getTypeIcon() {
+        return 'T';
+    }
+
```
natashatanyt, line 14: Again, consistency - usually your last closing brace does not have an extra empty line above it

src/main/java/Duke.java
```diff
@@ -1,10 +1,84 @@
+
+import java.util.Scanner;
+import java.util.ArrayList;
+
 public class Duke {
+    static ArrayList<Task> list = new ArrayList<Task>();
```
natashatanyt: You can use a better name (according to prof) - I think you should use a better name that's more specific, such as tasksList or something like that

src/main/java/Duke.java
```diff
@@ -1,10 +1,84 @@
...
+                list.get(i).getTypeIcon(), list.get(i).getStatusIcon(), list.get(i).getDescription()));
+    }
     public static void main(String[] args) {
         String logo = " ____        _        \n"
                 + "|  _ \\ _   _| | _____ \n"
                 + "| | | | | | | |/ / _ \\\n"
                 + "| |_| | |_| |   <  __/\n"
                 + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
+        System.out.println("Hello I am Baymax your personal life assistant \n" + logo);
```
natashatanyt: Not really a coding standard violation, but if your assistant is called Baymax, why is the ASCII art DUKE?

natashatanyt (commented):
Not sure where to leave a general comment so I'll put it here:
All in all, good job! It's just very minor issues w/ your code.

It's very good that you followed the camel casing your methods.

Keep it up!

src/main/java/Duke.java
```diff
@@ -13,37 +19,56 @@ public static void main(String[] args) {
         System.out.println("What can I do for you today?");
         String line;
         Scanner in = new Scanner(System.in);
-        ArrayList<Task> list = new ArrayList<Task>();
+
+
         while (true) {
```
Reynold-SL: Perhaps consider using a switch statement to make the code more readable?

src/main/java/Event.java
```diff
@@ -0,0 +1,36 @@
+public class Event extends Task {
+    String from;
```
Reynold-SL, line 2:
```suggestion
    protected String from;
```
perhaps you can consider making this variable protected to prevent accidental changes?

src/main/java/Deadline.java
```diff
@@ -0,0 +1,27 @@
...
+    public String getBy() {
+        return by;
+    }
+
+    public void setBy(String by) {
+        this.by = by;
+    }
+
+    Deadline(String description, String by) {
```
Reynold-SL:
```suggestion
public Deadline(String description, String by) {
```
Perhaps you can amend the method by adding a public for accessibility?

src/main/java/Duke.java
```diff
@@ -13,37 +19,56 @@ public static void main(String[] args) {
...
             line = in.nextLine();
             String arr[] = line.split(" ", 2);
             if (line.equals("bye")) {
                 System.out.println("Bye see you again!");
                 break;
             } else if (line.equals("list")) {
-                System.out.println("Here's your plan for a productive day!");
+                if (list.size() == 0) {
```
Reynold-SL: perhaps you can create a method to check for the size to prevent nested loops?

src/main/java/Duke.java
```diff
@@ -13,37 +19,56 @@ public static void main(String[] args) {
...
+                        list.get(i).getTypeIcon(), list.get(i).getStatusIcon(), list.get(i).getDescription()));
+
+            } else if (arr[0].equals("todo")) {
+                addTasks(new Todo(arr[1]));
+                System.out.println("Added!");
+                System.out.println(String.format(" [%s] [%s] %s", 'T', " ", arr[1]));
+                System.out.println("Now you have " + list.size() + " tasks in your list");
+            } else if (arr[0].equals("deadline")) {
+                String[] parts = arr[1].split("/");
+                addTasks(new Deadline(parts[0], parts[1]));
```
Reynold-SL:
```suggestions
Deadline[] deadline = new Deadline(parts[0], parts[1]);
addTasks(deadline);
```
perhaps you should create a variable first before adding to the method for better readability?

src/main/java/Deadline.java
```diff
@@ -0,0 +1,33 @@
+public class Deadline extends Task {
+    private String by;
+
+
```
kerct:
```suggestion
```

src/main/java/Duke.java
```diff
@@ -1,10 +1,146 @@
+
+import java.util.Objects;
+import java.util.Scanner;
+import java.util.ArrayList;
+
 public class Duke {
-    public static void main(String[] args) {
+    static ArrayList<Task> tasksList = new ArrayList<Task>();
+
+    private static void addTasks(Task a) {
```
kerct: It is better to name the variable (a in this case) that is reflective of what it is used for. In this case, perhaps you can rename 'a' to 'task'.

kerct: Also, since you are only adding 1 task, a better function name would be "addTask"

src/main/java/Duke.java
```diff
@@ -1,10 +1,146 @@
...
+            DukeException.IncompleteInputException {
+
+        if ((!Objects.equals(input[0], "bye"))
+                && (!Objects.equals(input[0], "list"))
+                && (!Objects.equals(input[0], "todo"))
+                && (!Objects.equals(input[0], "event"))
+                && (!Objects.equals(input[0], "deadline"))
+                && (!Objects.equals(input[0], "mark"))
+                && (!Objects.equals(input[0], "unmark"))
+                && (!Objects.equals(input[0], "delete"))) {
```
kerct: Since these are Strings, perhaps you can use the .equals method from the String class rather than the one from the Objects class

src/main/java/Duke.java
```diff
@@ -1,10 +1,146 @@
...
+
+    public static void main(String[] args) {
+        printIntro();
+        FileWriterClass.loadFromFile();
+        String line;
+        Scanner in = new Scanner(System.in);
+
+        while (true) {
+            line = in.nextLine();
+            String arr[] = line.split(" ", 2);
```
kerct: Perhaps you can rename this to something like input

src/main/java/FileWriterClass.java
```diff
@@ -0,0 +1,99 @@
+
+import java.io.*;
```
kerct: It's recommended to list the imported classes explicity rather than using the wildcard

src/main/java/Todo.java
```diff
@@ -0,0 +1,15 @@
+public class Todo extends Task {
+
+    public Todo(String description) {
+        super(description);
+       // System.out.println(String.format(" [%s] [%s] %s ", 'T', " ", description));
+
+
```
kerct, line 7: Perhaps you can remove this commented code which is not used; you can always check the revision history if you need it again

kerct (commented):
Good effort, but perhaps you can be more intentional in your naming!

PR #34, opened 2023-01-27
[L0Z1K] iP

L0Z1K:


src/main/java/Orca.java
```diff
@@ -0,0 +1,166 @@
...
+
+public class Orca {
+    static final String LOGO = "     _______  ______    _______  _______ \n"
+            + "    |       ||    _ |  |       ||   _   | \n"
+            + "    |   _   ||   | ||  |       ||  |_|  | \n"
+            + "    |  | |  ||   |_||_ |       ||       | \n"
+            + "    |  |_|  ||    __  ||      _||       | \n"
+            + "    |       ||   |  | ||     |_ |   _   | \n"
+            + "    |_______||___|  |_||_______||__| |__| \n";
+    static final int MAX_TASKS = 100;
```
superkaiba: I like how you named your constants in all capital letters with underscores as spaces

src/main/java/Orca.java
```diff
@@ -0,0 +1,166 @@
...
+
+    public static void printLatestAddedTask() {
+        System.out.println("    --------------------------------------------------");
+        System.out.println("    Got it. I've added this task:");
+        System.out.println("      " + tasks[taskCount - 1]);
+        System.out.println("    Now you have " + taskCount + " tasks in the list.");
+        System.out.println("    --------------------------------------------------\n");
+    }
+
+    public static void findCommandType() {
```
superkaiba: I like how you used camel case for all your method names

src/main/java/Orca.java
```diff
@@ -0,0 +1,166 @@
...
+    public static void finishProcess() {
+        in.close();
+    }
+
+    public static void main(String[] args) {
+        printGreetingMessage();
+        runOrca();
+        printByeMessage();
+        finishProcess();
+    }
```
superkaiba: Perhaps you could add some comments to clarify how your code works?

src/main/java/Orca.java
```diff
@@ -0,0 +1,166 @@
...
+        } else if (userInput.startsWith("unmark")) {
+            commandType = CommandType.UNMARK;
+        } else if (userInput.startsWith("todo")) {
+            commandType = CommandType.TODO;
+        } else if (userInput.startsWith("deadline")) {
+            commandType = CommandType.DEADLINE;
+        } else if (userInput.startsWith("event")) {
+            commandType = CommandType.EVENT;
+        } else {
+            commandType = CommandType.UNKNOWN;
```
superkaiba: I like how you followed the coding standards for this if/else if statement.

src/main/java/Orca.java
```diff
@@ -0,0 +1,166 @@
...
+    public static void printLatestAddedTask() {
+        System.out.println("    --------------------------------------------------");
+        System.out.println("    Got it. I've added this task:");
+        System.out.println("      " + tasks[taskCount - 1]);
+        System.out.println("    Now you have " + taskCount + " tasks in the list.");
+        System.out.println("    --------------------------------------------------\n");
+    }
+
+    public static void findCommandType() {
+        if (userInput.equals("bye")) {
```
superkaiba: Perhaps this if/else if statement could be replaced by a switch/case statement?

superkaiba (commented):
Really great!

src/main/java/Orca.java
L0Z1K: I will add some comments. Thank you.

src/main/java/Orca.java
L0Z1K: I used some `startswith()` so I don't think really I can do that.

src/main/java/Event.java
```diff
@@ -0,0 +1,18 @@
+public class Event extends Task {
+    protected String from;
+    protected String to;
+
+    Event(String userInput, int startIdx) {
+        int fromIndex = userInput.indexOf(" /from ");
```
manushridiv: Code quality is good and the variables are named appropropriately. Indentation is all good.

src/main/java/test.java
```diff
@@ -0,0 +1,10 @@
+import orca.Orca;
+
+public class test {
+    public static void main(String[] args) {
+        Orca o = new Orca();
```
qianz-z, line 5: Maybe you can have a better naming for your object.

src/main/java/test.java
```diff
@@ -0,0 +1,10 @@
...
+        o.printGreetingMessage();
+        o.runOrca();
+        o.finishProcess();
```
qianz-z: Good practice of SLAP!

src/orca/Deadline.java
```diff
@@ -0,0 +1,30 @@
+package orca;
+
+public class Deadline extends Task {
+    protected String by;
+
+    Deadline(String userInput, int startIdx) throws OrcaException {
+        try {
+            String[] deadline = userInput.substring(startIdx).split(" /by ");
+            this.description = deadline[0];
+            this.by = deadline[1];
```
qianz-z, line 10: Do take note of magic numbers as there are quite a number of them appearing in the rest of your code! You can try to use a named constant makes the code easier to understand because the name tells us more about the meaning of the number.

src/orca/Orca.java
```diff
@@ -0,0 +1,267 @@
...
+    static String userInput = "";
+    static CommandType commandType;
+    static int taskNo;
+    static Task newTask;
+
+    public Orca() {
+        File f = new File(FILE_PATH);
+        if (!f.exists()) {
+            f.getParentFile().mkdirs();
+        } else {
```
qianz-z: Try to avoid deep nesting / arrowhead style code as it is difficult to read and debug in the future.

src/orca/Orca.java
```diff
@@ -0,0 +1,267 @@
...
+
+    public static boolean isInputAvailable() {
+        return in.hasNextLine();
+    }
+
+    public static void readUserInput() {
+        userInput = in.nextLine();
+    }
+
+    public static void markTask(int taskNo) throws OrcaException {
```
qianz-z: Instead of `taskNo`, maybe you can consider renaming it as `taskNumber` so that it is more readable.

src/orca/Orca.java
```diff
@@ -0,0 +1,267 @@
...
+        try {
+            tasks.get(taskNo - 1).setDone(true);
+        } catch (NullPointerException e) {
+            throw new OrcaException("There is no task with this number.");
+        } catch (ArrayIndexOutOfBoundsException e) {
+            throw new OrcaException("There is no task with this number.");
+        }
+        System.out.println("    --------------------------------------------------");
```
qianz-z: You can consider removing the spaces at the front, or replace it by creating another constant so that it is more uniform.

qianz-z (commented):
Good effort in completing your iP on time! It is generally well done with some improvements to be made :>

PR #39, opened 2023-01-28
[Matthew Liu Zhen Jie] iP

matthew-liu-zhenjie:


src/main/java/Duke.java
```diff
@@ -1,10 +1,50 @@
...
     public static void main(String[] args) {
         String logo = " ____        _        \n"
                 + "|  _ \\ _   _| | _____ \n"
                 + "| | | | | | | |/ / _ \\\n"
                 + "| |_| | |_| |   <  __/\n"
                 + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
+        //System.out.println("Hello from\n" + logo);
+        printGreeting();
+        boolean byeFlag = false;
```
khooyourun: Better to name Boolean variables in a way that sounds like a boolean

src/main/java/Duke.java
```diff
@@ -1,10 +1,50 @@
...
+        printSeparator();
+        System.out.println("Bye. Hope to see you again soon!");
+        printSeparator();
+    }
+
+    public static void printEcho(String input) {
+        printSeparator();
+        System.out.println("added: " + input);
+        printSeparator();
+    }
```
khooyourun: I like the naming of methods here. It makes it clear what each method does.

src/main/java/TaskManager.java
```diff
@@ -0,0 +1,107 @@
...
+        } else if (input.contains("todo")){
+            generateToDo(input);
+        } else if (input.contains("deadline")){
+            generateDeadline(input);
+        } else if (input.contains("event")){
+            generateEvent(input);
+        } else{
+            printHelp();
+        }
+    }
```
khooyourun: I like the usage of else if statements instead of nested if statements. Makes it easy to understand.

src/main/java/TaskManager.java
```diff
@@ -0,0 +1,107 @@
...
+
+    public void markTaskAsDone(int taskIndex){
+        taskList[taskIndex-1].markAsDone();
+    }
+    public void markTaskAsUndone(int taskIndex){
+        taskList[taskIndex-1].markAsUndone();
+    }
+    public void markTask(String input) {
+        int whitespaceIndex = input.indexOf(" ");
```
khooyourun: Preferably name whitespaceIndex as whiteSpaceIndex with a capital S (camelCase).

src/main/java/TaskManager.java
```diff
@@ -0,0 +1,107 @@
...
+            generateDeadline(input);
+        } else if (input.contains("event")){
+            generateEvent(input);
+        } else{
+            printHelp();
+        }
+    }
+
+    private void generateToDo(String input) {
+        String job = input.substring(5);
```
khooyourun: What does the 5 in input.substring(5) represent? Same question for the other input.substring()

khooyourun (commented):
Overall the code is easy to read as it uses well named modules and variables.

src/main/java/TaskManager.java
matthew-liu-zhenjie: Hey khooyourun, thanks for the comment. However, whitespace is a single word

src/main/java/TaskManager.java
matthew-liu-zhenjie: Noted, I'll comment as required for clarity. This creates a substring from input[5] onwards.

PR #44, opened 2023-01-28
[Jeraldchen] iP

Jeraldchen:


src/main/java/Duke.java
```diff
@@ -1,10 +1,69 @@
...
     public static void main(String[] args) {
-        String logo = " ____        _        \n"
-                + "|  _ \\ _   _| | _____ \n"
-                + "| | | | | | | |/ / _ \\\n"
-                + "| |_| | |_| |   <  __/\n"
-                + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
+        String line;
+        String lowerCaseLine;
+        Task[] textList = new Task[100]; //store in array of Tasks to track both description and mark status
```
cleoncheng2000: Perhaps you can create a variable for the array size 100? So that it will be easier to change in the future.

src/main/java/Duke.java
```diff
@@ -1,10 +1,69 @@
...
+                printLine();
+                return;
+            } else if (lowerCaseLine.equals("list")) { //if input is list
+                System.out.println("Here are the tasks in your list: ");
+                for (int i = 0; i < textListCount; ++i) {
+                    String listLine = i+1 + ". " + textList[i].toString();
+                    System.out.println(listLine);
+                }
+                printLine();
+            } else if (lowerCaseLine.contains("mark") && !lowerCaseLine.contains("unmark")) { //mark is in unmark so need to check like this to avoid unmark being detected as mark
```
cleoncheng2000: This is a long else-if line. Perhaps you can use a more intuitive method here?

src/main/java/Duke.java
```diff
@@ -1,10 +1,69 @@
...
+                System.out.println(" " + textList[indexOfTask].toString());
+                printLine();
+            } else if (lowerCaseLine.contains("unmark")) {
+                int indexOfTask = Integer.parseInt(lowerCaseLine.substring(7)) - 1;
+                textList[indexOfTask].unmarkAsDone();
+                System.out.println("OK, I've marked this task as not done yet: ");
+                System.out.println(" " + textList[indexOfTask].toString());
+                printLine();
+            }
+            else { //if input is any other text
```
cleoncheng2000: Nested if-else statements in else statement. Perhaps you can extract this out and separate them?

src/main/java/Duke.java
cleoncheng2000: Perhaps you can use a more intuitive variable name for your list in the context of Duke?

src/main/java/Duke.java
```diff
@@ -1,10 +1,69 @@
...
-        String logo = " ____        _        \n"
-                + "|  _ \\ _   _| | _____ \n"
-                + "| | | | | | | |/ / _ \\\n"
-                + "| |_| | |_| |   <  __/\n"
-                + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
+        String line;
+        String lowerCaseLine;
+        Task[] textList = new Task[100]; //store in array of Tasks to track both description and mark status
+        int textListCount = 0;
```
cleoncheng2000: Perhaps you can use a more intuitive variable name here?

cleoncheng2000 (commented):
Overall looks good!

src/main/java/Event.java
```diff
@@ -0,0 +1,38 @@
...
+import java.time.LocalDateTime;
+
+/**
+    * Represents an event task.
+ */
+public class Event extends Task{
+    protected String from;
+    protected String to;
+    protected String from_copy;
+    protected String to_copy;
```
chinhan99, line 12: Do check your variable naming convention!

src/main/java/Parser.java
```diff
@@ -0,0 +1,111 @@
...
+                Ui.printFindMessage(TaskList.findTask(keyword));
+            } catch (StringIndexOutOfBoundsException e) {
+                Ui.printEmptyDescription();
+            }
+            break;
+        default:
+            Ui.printInvalidCommand();
+            break;
+        }
+    }
```
chinhan99, line 102: This method is too long. Do consider abstracting out the submethods for readability.

src/main/java/Storage.java
```diff
@@ -0,0 +1,95 @@
...
+                    textList.add(deadline);
+                } else if (taskType.equals("E")) {
+                    String[] timeArray = lineArray[3].trim().split("to");
+                    String from = timeArray[0].trim();
+                    String to = timeArray[1].trim();
+                    Event event = new Event(description, from, to);
+                    if (isDone) {
+                        event.markAsDone();
+                    }
+                    textList.add(event);
```
chinhan99, line 87: lengthy arrowhead code observed within here. Do consider abstracting out the submethods. In addition , values such as "E" could be replaced with a constant too.

chinhan99 (commented):
Good work on this IP.  Do consider abstracting out methods from those lengthy methods for reasons. Some of your magic values can also be replaced by constants too( also note to name the constants meaningfully)!

PR #57, opened 2023-01-28
[Oong Jin Rong Jared] iP

jaredoong:


src/main/java/Deadline.java
```diff
@@ -0,0 +1,12 @@
...
+    public Deadline(String description, String endDate) {
+        super(description);
+        this.endDate = endDate;
+    }
+
+    @Override
+    public String toString() {
+        String taskOutput = super.toString();
+        return String.format("[D]%s (by: %s)", taskOutput, endDate);
+    }
```
JangusRoundstone: This is an efficient way to output the string, good job!

src/main/java/Duke.java
```diff
@@ -1,10 +1,234 @@
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Scanner;
+
 public class Duke {
+
+    public static boolean isCompleted = false;
```
JangusRoundstone: Perhaps this variable could be initialised closer to the main function, so that the readers can better follow your code. This makes the happy path more prominent.

src/main/java/Duke.java
```diff
@@ -1,10 +1,234 @@
...
+    private static final List<Task> userList = new ArrayList<>();
+
+    public static void welcomeMessage() {
+        separator();
+        printMessage("Hello! I'm Duke");
+        printMessage("What can I do for you?");
+        separator();
```
JangusRoundstone: Great job writing the separator() function, could be really useful later on in iP!

src/main/java/Duke.java
```diff
@@ -1,10 +1,234 @@
...
+    }
+
+    public static void checkInput(String inputMessage) {
+        String[] message = inputMessage.split(" ");
+        if (message.length == 1 && message[0].equalsIgnoreCase("bye")) {
+            isCompleted = true;
+            return;
+        }
+
+        if (message.length == 1 && message[0].equalsIgnoreCase("list")) {
```
JangusRoundstone: Good job using the equalsIgnoreCase() function!

src/main/java/Duke.java
```diff
@@ -1,10 +1,234 @@
...
+    public static void printMessage(String message) {
+        String outputMessage = String.format("| %-57s|", message);
+        System.out.println(outputMessage);
+    }
+
+    public static void checkInput(String inputMessage) {
+        String[] message = inputMessage.split(" ");
+        if (message.length == 1 && message[0].equalsIgnoreCase("bye")) {
+            isCompleted = true;
+            return;
```
JangusRoundstone: Is it better to keep checking for message.length() > 1, or in this case it is better to check for "bye" or "list" at the beginning? I am unsure also, feel free to discuss:)

JangusRoundstone (commented):
LGTM, with one question that is debatable. Good job overall!

src/main/java/Task.java
```diff
@@ -1,9 +1,13 @@
 public class Task {
     protected String description;
+    protected String startDate;
+    protected String endDate;
     protected boolean isDone;
```
Aung-Phone-Naing: Good use of protected access-modifier to ensure no other classes can directly access and modify it.

src/main/java/Deadline.java
Aung-Phone-Naing: Good practice to use inheritance from Task class and then overriding for Deadline and Todo classes respectively for printing different types of outputs.

Aung-Phone-Naing (commented):
Overall I think you did a great job with the practices!

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,46 @@
...
+    public Task(String description) {
+        this.description = description;
+        this.startDate = "";
+        this.endDate = "";
+        this.isDone = false;
+    }
+
+    public String getStatusIcon() {
+        return ((this.isDone) ? "X" : " ");
+    }
```
theopin: Try to replace these string literals with named constants to better reflect their context

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,452 @@
...
+
+    public static void main(String[] args) {
+        try {
+            startDuke();
+        } catch (IOException error) {
+            printMessage("Unable to load/create database file! Printing error stack trace...");
+            error.printStackTrace();
+        }
+
+        String line;
```
theopin: try catch should have a finally clause to run code that needs to run regardless of the outcome of try-catch

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,452 @@
...
+            BufferedReader bufferedReader = new BufferedReader(fileReader);
+            String fileContent = "";
+            String line;
+            while ((line = bufferedReader.readLine()) != null) {
+                fileContent += line + "\n";
+            }
+
+            loadTaskList(fileContent);
+            printMessage("Done loading previous task list:");
+            String outputMessage = String.format("Loaded a total of %d tasks!", userList.size());
```
theopin: perhaps consider extracting related sub-behaviours such as this section into another method to simplify readability of startDuke.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,452 @@
...
+        } catch (IndexOutOfBoundsException error) {
+            String errorMessage = "Out of bounds value provided.";
+            String errorMessageEcho = String.format("List only has %d items, you entered %d!",
+                    userList.size(), taskNumber);
+            printMessage(errorMessage);
+            printMessage(errorMessageEcho);
+        } finally {
+            separator();
+        }
+    }
```
theopin: method seems quite long, perhaps consider refactoring as suggested above

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,452 @@
...
+            addTodo(cleanInput);
+        } else if (message[0].equalsIgnoreCase("deadline")) {
+            addDeadline(message);
+        } else if (message[0].equalsIgnoreCase("event")) {
+            addEvent(message);
+        } else if (message[0].equalsIgnoreCase("delete")) {
+            deleteTask(message);
+        } else {
+            unknownCommand();
+        }
```
theopin: Good level of abstraction

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,452 @@
...
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Scanner;
+
+public class Duke {
+
+    static int BOX_WIDTH = 100;
```
theopin: Perhaps also declare it as a final variable since the naming convention seems to suggest it is a constant?

theopin (commented):
Good job on your implementation of Duke so far! Some considerations to help polish your code:
- Magic literals: Try to avoid putting values (eg. "deadline", 123) as your function parameters, but rather declare them as named constants to so that readers better understand what the code is for
- Long Methods: Some of your methods seems to nearing/too long, which can make it difficult for refactors especially those that arise from logic issues. Try to identify common behaviours and extract them out into smaller methods so as to ease readability.

PR #68, opened 2023-01-29
[Abijith Ram] iP

TopGun2001:


src/main/java/Rolex.java
```diff
@@ -1,89 +1,115 @@
...
+    }
+
+    public static void InputIsDeadline(String UserInput){
+        int indexOfBy = UserInput.indexOf("/by");
+        String deadlineName = UserInput.substring(9,indexOfBy-1);
+        String by = UserInput.substring(indexOfBy+3);
+        task[taskCount] = new Deadline(deadlineName, by);
+        taskCount++;
+        Add_Print_Task();
+    }
```
0nandon: Good job for separating your code in to several functions.

src/main/java/Rolex.java
```diff
@@ -1,89 +1,115 @@
...
+        int indexNum = 1;
+        System.out.println("Here are the tasks in your list:");
+        for(int i = 0; i<taskCount; i++){
+            System.out.println(indexNum + ". " + task[i]);
+            indexNum++;
+        }
+        printLines();
+    }
+
+    public static void InputIsMark(String UserInput){
```
0nandon:
```suggestion
public static void inputIsMark(String userInput){
```

0nandon: When you declare function or input, It is more better to user lower case in the first letter of the name.

0nandon (commented):
I think code is quite clean, especially for separating every code module into multiple functions. But in my knowledge, first letter of the function, input, variables, etc is better to be in lower case. Otherwise, I think the code is well structured.

src/main/java/Rolex.java
```diff
@@ -1,89 +1,115 @@
...
+        int index = Integer.parseInt(UserInput.substring(5));
+        task[index-1].MarkTask();
+    }
+
+    public static void InputIsUnmark(String UserInput){
+        int index = Integer.parseInt(UserInput.substring(7));
+        task[index-1].unMarkTask();
+    }
+
+    public static void Add_Print_Task(){
```
0nandon:
```suggestion
public static void addPrintTask(){
```

0nandon (commented):
In my knowledge, underline should be only used in the test methods.

src/main/java/Rolex.java
```diff
@@ -1,89 +1,115 @@
 import java.util.Scanner;
 public class Rolex {
```
0nandon:
```suggestion
import java.util.Scanner;

public class Rolex {
```

0nandon (commented):
It is better to put single line between the import packages and class you declared.

src/main/java/Rolex.java
```diff
@@ -1,89 +1,115 @@
...
-                String eventName = UserInput.substring(6,indexOfFrom-1);
-                String startTime = UserInput.substring(indexOfFrom+6,indexOfTo-1);
-                String endTime = UserInput.substring(indexOfTo+4);
-                task[taskCount] = new Event(eventName, startTime, endTime);
-                taskCount++;
-                System.out.println("Got it. I've added this task:");
-                System.out.println(task[taskCount-1]);
-                System.out.println("\nNow you have " + taskCount + " tasks in the list.");
-                printLines();
+                InputIsEvent(UserInput);
```
0nandon: 'switch' is more better than if-else statement in this case since there are various cases you want to separate.

0nandon (commented):
If there are numerous cases you want to separate, 'switch' could be your answer rather than if-else.

src/main/java/Rolex.java
```diff
@@ -1,89 +1,115 @@
...
+                InputIsList();
             }
-
             else if(UserInput.startsWith("mark")){
-                int index = Integer.parseInt(UserInput.substring(5));
-                task[index-1].MarkTask();
+                InputIsMark(UserInput);
             }
-
             else if(UserInput.startsWith("unmark")){
```
0nandon:
```suggestion
}else if(UserInput.startsWith("unmark")){
```

0nandon (commented):
format of the if-else statement should be like below:

if(...){
    ...
} else if(...){
    ...
}

TopGun2001:
>

src/main/java/Rolex.java
TopGun2001: I didn't know that it is suggested to make the first letters lowercase. I will change them!

src/main/java/Rolex.java
TopGun2001: Got it!

src/main/java/Rolex.java
TopGun2001: Actually, I have changed the coding standard and quality after doing my level-4. In that, most of the lines u see in red are inside separate functions. They look short and concise now. Thanks for your feedback!

src/main/java/Rolex.java
TopGun2001: I didn't know this until reading the coding standards. Will make the necessary changes!

src/main/java/tasks/Task.java
```diff
@@ -0,0 +1,50 @@
...
+        } else{
+            System.out.println("tasks.Task.tasks.Task is already unmarked.");
+        }
+        printLines();
+    }
+
+    public String taskStatus(){
+        return (isDone ? "X" : " ");
+    }
+
```
theopin: Try to avoid using string literals such as these by converting them to named constants. This will help in readability of code

src/main/java/tasks/Event.java
```diff
@@ -0,0 +1,17 @@
...
+        super(taskName);
+        this.startTime = startTime;
+        this.endTime = endTime;
+    }
+
+    public String toString(){
+        return "[E]" + super.toString()+ " (from: " + startTime + " to: " + endTime + ")";
+    }
+
+} // tasks.Task.tasks.Event class ends here
```
theopin: Any purpose for this style of comment?

src/main/java/exception/RolexException.java
```diff
@@ -0,0 +1,53 @@
...
+        } else if (userInput.startsWith("mark")){
+            runMarkError();
+        } else if (userInput.startsWith("unmark")){
+            runUnmarkError();
+        } else if(userInput.startsWith("list")){
+            runListError();
+        } else{
+            System.out.println("OOPS!! I cannot understand the input! ☹ ");
+        }
+        printLines();
```
theopin: Good level of abstraction

src/main/java/Rolex.java
```diff
@@ -0,0 +1,140 @@
...
+        if(indexOfFrom == -1 || indexOfTo == -1){
+            RolexException.detectError(userInput);
+
+        } else{
+            String eventName = userInput.substring(6,indexOfFrom-1);
+            String startTime = userInput.substring(indexOfFrom+6,indexOfTo-1);
+            String endTime = userInput.substring(indexOfTo+4);
+            task[taskCount] = new Event(eventName, startTime, endTime);
+            taskCount++;
+            addPrintTask();
```
theopin: Seems similar to 87-89 and  74-76, perhaps you could combine them in 1 method since apart from the type of task being instantiated, they exhibit similar behaviour

src/main/java/Rolex.java
```diff
@@ -0,0 +1,140 @@
...
+    public static void inputIsMark(String userInput){
+        int index = Integer.parseInt(userInput.substring(5));
+        if(index>0 && index<=taskCount) {
+            task[index-1].MarkTask();
+        } else{
+            RolexException.detectError(userInput);
+        }
+    }
+
+    public static void inputIsUnmark(String userInput){
```
theopin: Try to rephrase your method names to act as verbs rather than boolean conditionals (eg. handleUnmarkInput)

theopin (commented):
Good implementation of Duke! Some notes to polish your code:
- Magic literals: Try to avoid putting strings/ints into function inputs by converting them into named constants
- Variable naming: The methods handling inputs seems confusing, try to rename them using the style suggested (verb + context to be taken care of)

src/main/java/Rolex.java
TopGun2001: I have seen all the reviews from you and will format accordingly. Thanks @theopin !

PR #74, opened 2023-01-29
Level-3 done, PR for IP week4

hyperbola-bear:
trying to create pull request by order of CS2113 for IP week4 after completing level-3


src/main/java/Duke.java
```diff
@@ -1,3 +1,8 @@
+import java.util.Scanner;
+import java.util.ArrayList;
```
nikkiDEEE: I like how you didnt just use `java.util.*`!

src/main/java/Duke.java
```diff
@@ -6,5 +11,44 @@ public static void main(String[] args) {
                 + "| |_| | |_| |   <  __/\n"
                 + "|____/ \\__,_|_|\\_\\___|\n";
         System.out.println("Hello from\n" + logo);
+        System.out.println("Hello! I'm Duke");
+        System.out.println("What can I do for you?");
+        Scanner in = new Scanner(System.in);
+        String input;
+        ArrayList<String> items = new ArrayList<String>();
```
nikkiDEEE: I like how you used a resizable array!

src/main/java/Duke.java
```diff
@@ -6,5 +11,44 @@ public static void main(String[] args) {
...
+        ArrayList<Boolean> marked = new ArrayList<Boolean>();
+        
+        while((input = in.nextLine()) != "") {
+            if (input.equalsIgnoreCase("bye")) {
+                System.out.println("Bye. Hope to see you again soon!");
+                in.close();
+                return;
+            }
+            else if (input.equals("list")) {
+
```
nikkiDEEE: Perhaps this empty nextline can be omitted?

src/main/java/Duke.java
```diff
@@ -6,5 +11,44 @@ public static void main(String[] args) {
...
+            if (input.equalsIgnoreCase("bye")) {
+                System.out.println("Bye. Hope to see you again soon!");
+                in.close();
+                return;
+            }
+            else if (input.equals("list")) {
+                
+                for(int i = 0; i < items.size(); i++) {
+                    System.out.println(i+1 +". " +  (marked.get(i) ? "[X] " : "[ ] ")   + items.get(i));
+
```
nikkiDEEE: Perhaps this nextline comment can be removed to adhere to the coding standards?

src/main/java/Duke.java
```diff
@@ -6,5 +11,44 @@ public static void main(String[] args) {
                 + "| |_| | |_| |   <  __/\n"
                 + "|____/ \\__,_|_|\\_\\___|\n";
         System.out.println("Hello from\n" + logo);
+        System.out.println("Hello! I'm Duke");
+        System.out.println("What can I do for you?");
+        Scanner in = new Scanner(System.in);
+        String input;
+        ArrayList<String> items = new ArrayList<String>();
+        ArrayList<Boolean> marked = new ArrayList<Boolean>();
```
nikkiDEEE: Perhaps you could use the array variable name `isMarked` so it sounds like a boolean?

nikkiDEEE (changes requested):
Overall, the codeing standard indentation lengths have been met but there are some minor issues with the empty lines which can be removed to make the overall code slightly cleaner. Keep up the good work!

src/main/java/Duke.java
```diff
@@ -6,5 +11,44 @@ public static void main(String[] args) {
...
+                String[] strArray = input.split(" ");
+                int num = Integer.parseInt(strArray[1]);
+                marked.set(num-1, true);
+                System.out.println("Nice! I've marked this task as done:\n" + "[X] " + items.get(num-1));
+            }
+            else if (input.startsWith("unmark")) {
+                String[] strArray = input.split(" ");
+                int num = Integer.parseInt(strArray[1]);
+                marked.set(num-1, false);
+                System.out.println("Ok, I've marked this task as not done yet:\n" + "[ ] " + items.get(num-1));
```
coregano: I realised that your "mark" and "unmark" commands do not check if the string after whitespace is convertable to type Integer. I believe that this may bring up an error if the user were to input in a command "mark qwerty", as "qwerty" cannot be passed into the method Integer.parseInt(). I recommend using a method to check for the formatting of the command first, such as String.matches() or even a try-catch statement. 
I believe that this would greatly strengthen the ability of your code to handle such edge cases.

src/main/java/Duke.java
```diff
@@ -6,5 +11,44 @@ public static void main(String[] args) {
...
+            }
+            else if (input.equals("list")) {
+                
+                for(int i = 0; i < items.size(); i++) {
+                    System.out.println(i+1 +". " +  (marked.get(i) ? "[X] " : "[ ] ")   + items.get(i));
+                    
+                }
+            }
+            else if (input.startsWith("mark")) {
+                String[] strArray = input.split(" ");
```
coregano: I recommend executing the command String.trim() before using String.split(), as it would account for cases where the user has inputted whitespace(s) before the actual non-whitespace text.

src/main/java/Duke.java
```diff
@@ -6,5 +11,44 @@ public static void main(String[] args) {
...
         System.out.println("Hello from\n" + logo);
+        System.out.println("Hello! I'm Duke");
+        System.out.println("What can I do for you?");
+        Scanner in = new Scanner(System.in);
+        String input;
+        ArrayList<String> items = new ArrayList<String>();
+        ArrayList<Boolean> marked = new ArrayList<Boolean>();
+        
+        while((input = in.nextLine()) != "") {
+            if (input.equalsIgnoreCase("bye")) {
```
coregano: I recommend standardising between input.equals() and input.equalsIgnoreCase() for bye and other commands, so as to make the code easier to understand and debug incase an error arises during testing.

src/main/java/Duke.java
coregano: Using two resizable arrays may affect runtime and make it harder to code as there needs to be constant linkage between the two items in each array. Although it may seem simple now, this will become much harder to maintain when you have more methods being implemented into your code in the future, such as deleting and moving items along the list. Hence, I believe that creating a separate class Task may benefit the code and yourself greatly.

coregano (commented):
Well-written and clean code with high readability. 
Could be improved on the handling of edge cases.

src/main/java/Deadline.java
```diff
@@ -0,0 +1,14 @@
+public class Deadline extends Task {
+    public String description;
+
+
+    public Deadline(String input) {
+        String[] strArray = input.split("/");
```
aditichadha1310: Try to avoid the usage of Magic String/literals.

src/main/java/Event.java
```diff
@@ -0,0 +1,11 @@
+public class Event extends Task{
+    public String description;
+    public Event(String input) {
+        String[] strArray = input.split("/");
+        this.description = strArray[0].replace("deadline ", "") + ( "(" + strArray[1].replace("from", "from:")) + (strArray[2].replace("to", "to:") + ")");
```
aditichadha1310: It is recommended to have at maximum 120 characters on one line (hard limit). Try to divide it into two lines to keep the number of characters in each line less than 120.

src/main/java/Task.java
```diff
@@ -0,0 +1,40 @@
+import java.util.ArrayList;
+
+public class Task {
+    public enum TaskType {
+        TODO, DEADLINE, EVENT
```
aditichadha1310, line 6: Good job on using enumerations!

src/main/java/Duke.java
```diff
@@ -1,10 +1,57 @@
...
+               
+                new Deadline(input);
+            }
+            else if (input.startsWith("event")) {
+                
+                new Event(input);
+            }
+            else {
+                System.out.println("invalid input");
+            }
```
aditichadha1310: This is a suggestion:- To avoid the usage of Magic Strings - you can consider saving these strings ("event", "invalid input" etc. ) as some variables and use the variables to check for equality.

aditichadha1310 (commented):
Good job overall!

PR #78, opened 2023-01-29
[Hong Lin Shang] iP

honglinshang:


src/main/java/Duke.java
```diff
@@ -1,10 +1,205 @@
...
+            if (canStore()) {
+                addToDo(param);
+            }
+            break;
+        case COMMAND_DEADLINE_WORD:
+            if (canStore()) {
+                final String[] paramAndBy = processDeadline(param);
+                if (paramAndBy == null) {
+                    System.out.println(DIVIDER +
+                            "Please enter deadline as \"deadline [task] /by [date]\". \n" +
```
shawntangy: Could consider extracting this line out as a constant 'DEADLINE_INSTRUCTION'.

src/main/java/Duke.java
```diff
@@ -1,10 +1,205 @@
...
+            if (idx < 0 || idx >= counter) {
+                printErrorForIdx(counter);
+                return;
+            }
+            store[idx].setDone(true);
+            System.out.println(DIVIDER + "Nice!, I've marked this task as done:\n" +
+                    store[idx] + "\n" + DIVIDER);
+        } catch (NumberFormatException ex) {
+            printErrorForIdx(counter);
+        }
```
shawntangy: I really liked how you use a try and catch here!

src/main/java/Duke.java
```diff
@@ -1,10 +1,205 @@
...
+    }
+
+    private static void addToDo(String param) {
+        allTasks[counter] = new ToDo(param);
+        printAddMessage(allTasks[counter]);
+        counter++;
```
shawntangy: Maybe could do a static counter variable in the Task Class instead of Duke Class. Or is there any reason you chose to do it this way?

src/main/java/Event.java
```diff
@@ -0,0 +1,31 @@
...
+        return to;
+    }
+
+    public void setTo(String to) {
+        this.to = to;
+    }
+
+    @Override
+    public String toString() {
+        return "[E][" + getStatus() + "] " + description + " (from: " + from + " to: " + to + ")";
```
shawntangy: Maybe can extract out the '[E]' to be a constant too, also applies for deadline/todo [D]/[T]

shawntangy (commented):
Overall the code is very neat, well done :)

src/main/java/Duke.java
```diff
@@ -1,10 +1,205 @@
+import java.util.Scanner;
+
 public class Duke {
+    private static final String COMMAND_TODO_WORD = "todo";
+    private static final String COMMAND_DEADLINE_WORD = "deadline";
+    private static final String COMMAND_EVENT_WORD = "event";
+    private static final String COMMAND_MARK_WORD = "mark";
+    private static final String COMMAND_UNMARK_WORD = "unmark";
+    private static final String COMMAND_LIST_WORD = "list";
+    private static final String COMMAND_EXIT_WORD = "bye";
```
wenxin-c: Nice use of constants:D

src/main/java/Duke.java
```diff
@@ -1,10 +1,205 @@
...
-                + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
+        initTasks();
+        printWelcomeMessage();
+        while (true) {
+            String userCommand = in.nextLine();
+            final String[] commandAndParam = processCommand(userCommand);
+            String command = commandAndParam[0];
+            String param = commandAndParam[1];
+            executeCommand(command, param);
```
wenxin-c: Good abstraction into different command methods. It is neat and readable.

src/main/java/Duke.java
```diff
@@ -1,10 +1,205 @@
...
+        }
+    }
+
+    private static boolean canStore() {
+        if (counter < CAPACITY) {
+            return true;
+        }
+        printErrorForStorage();
+        return false;
+    }
```
wenxin-c: Nice check of the capacity of array to catch errors:D

src/main/java/Duke.java
```diff
@@ -1,10 +1,205 @@
...
+                return;
+            }
+            store[idx].setDone(false);
+            System.out.println(DIVIDER +
+                    "OK, I've marked this task as not done yet:\n" +
+                    "[ ] " + store[idx].getDescription() + "\n" +
+                    DIVIDER);
+        } catch (NumberFormatException ex) {
+            printErrorForIdx(counter);
+        }
```
wenxin-c: Good use of try catch for catch possible errors.

wenxin-c (commented):
Overall good work. To further improve the code quality and readability, the code can be further extracted into classes. Instead of writing all the command methods inside Duke class, they can be extracted into a command manager class. The task methods can also be extracted into a task manager class. As such, Duke class can be kept concise, increasing the readability of the code.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    private static void deleteTask(String param) {
+        int idx = Integer.parseInt(param) - 1;
+        if (idx < 0 || idx >= allTasks.size()) {
+            throw new NumberFormatException();
+        }
+        Output.printDeleted(allTasks.get(idx));
+        allTasks.remove(idx);
+
```
kerct:
```suggestion
```

src/main/java/duke/exception/EmptyTaskException.java
```diff
@@ -0,0 +1,4 @@
+package duke.exception;
+
+public class EmptyTaskException extends Throwable {
```
kerct: From the Javadoc https://docs.oracle.com/javase/7/docs/api/java/lang/Throwable.html, "The Throwable class is the superclass of all errors and exceptions in the Java language". In this case, since EmptyTaskException should be an Exception and not an Error, it is recommended to extend from the Exception class rather than Throwable, which includes both Exception and Error.

src/main/java/duke/task/Deadline.java
```diff
@@ -0,0 +1,29 @@
+package duke.task;
+
+public class Deadline extends Task {
+
+    public static final String DEADLINE_LABEL = "D";
```
kerct, line 16: If this variable doesn't need to be accessed outside of this class, it is probably better not to declare it as public

kerct (commented):
Code is well-organised, readable, and follows the coding standards 👍

PR #80, opened 2023-01-29
[Dadevchia] iP

DaDevChia:


src/main/java/Check.java
```diff
@@ -0,0 +1,72 @@
+public class Check {
```
clement559: Class name is slightly ambiguous.

src/main/java/Check.java
```diff
@@ -0,0 +1,72 @@
...
+     //Function to check if an input has mark as the first word
+     public static boolean isMark(String input){
+        if(input.length()>=4){
```
clement559: Input lengths appear to be magic numbers. Would be better to explain the different numbers used.

src/main/java/Duke.java
```diff
@@ -1,10 +1,137 @@
+import java.util.Scanner;  // Import the Scanner class
```
clement559: Line is obvious, comment can be omitted.

src/main/java/Duke.java
```diff
@@ -1,10 +1,137 @@
...
+        System.out.println("Hello! I'm Duke");
+        System.out.println("What can I do for you?");
+
+        //Get user input
+        Scanner userScan = new Scanner(System.in);  // Create scanner object
+        String userInput = userScan.nextLine();  // Get user input
+
+        //Check the arguments provided unless it is "bye" which quits the program
+        while( !(Check.isBye(userInput)) ){
+
```
clement559: Extra newline can be omitted.

src/main/java/Duke.java
```diff
@@ -1,10 +1,137 @@
+import java.util.Scanner;  // Import the Scanner class
 public class Duke {
+
+    //Initialize create a list of tasks
+    public static Task[] tasks = new Task[100];
+    public static int taskCount = 0;
+
     public static void main(String[] args) {
```
clement559: Main method is too long, can consider breaking up the different parts of the program into different methods.

src/main/java/Check.java
thomasjlalba: Would be good if there is spacing among the variables and operators, noticed the same thing for the other places too
```suggestion
        if (input.length() >= 4) {
```

thomasjlalba: This particular indentation seems to only be 3 spaces instead of the typical 4 spaces

src/main/java/Check.java
```diff
@@ -0,0 +1,72 @@
...
+            if("mark".equalsIgnoreCase(input.substring(0,4))){
```
thomasjlalba: Would be nice if there is a space between the bracket and if as such:
```suggestion
            if ("mark".equalsIgnoreCase(input.substring(0,4))) {
```

src/main/java/Task.java
```diff
@@ -0,0 +1,74 @@
...
+    public String toString(){
+        if(this.isDone){
+            return "[X] " + this.taskName;
+        }else{
+            return "[ ] " + this.taskName;
+        }
+    }
+}
+
+//Create a todo class that inherits from the task class
```
thomasjlalba: Would be nice to expand on the method's details using
```
/**
 * comment
 */
```

src/main/java/Duke.java
```diff
@@ -1,10 +1,137 @@
...
+            //If userInput is not any of the commands add the task to the list
+            else{
+                System.out.println("added: " + userInput); // Print user input
+                //add to task list
+                tasks[taskCount]=new Task(userInput);
+                taskCount++;
+                //Get user input again
+                userInput = userScan.nextLine();  
+            }
+
```
thomasjlalba: Perhaps you could remove this empty line

thomasjlalba (commented):
Overall the code is well written to the coding standards guidelines!

clement559 (commented):
Overall, code is fine, but lengthy methods at certain points. Splitting up the main code into segments will help improve readability!

src/main/java/Check.java
```diff
@@ -0,0 +1,82 @@
...
+                return true;
+            }
+        }
+        return false;
+    }
+
+    //Function to check if an input has unmark as the first word
+    public static boolean isUnmark(String input){
+        if(input.length()>=6){
```
wcwy: As identified by other reviewers, there should be a spaces in front and behind the bracket containing the if-condition.
Sample: 
```
if (isDone) {
    doCleanup();
}
```

Also, for the condition, if you are using operators such as `=, >, <, >=, <=, ==, !=, ||, && etc.`, try to put a space before and after the operator.
For example, we want to have `x != 123` instead of `x!=123`. 

Please check through your entire code base on this spacing issue. The same applies to other structure such as while loop. If you are not sure on where to put your spaces and where not, do take a look at the Java Coding Standard document.

src/main/java/Check.java
```diff
@@ -0,0 +1,82 @@
...
+            if("unmark".equalsIgnoreCase(input.substring(0,6))){
+                return true;
+            }
```
wcwy: Instead of having two if structure each for a condition, you can consider doing something like this:
```
boolean isValidInput = condition1 && condition2;
if (isValidInput) {
    // do something when input is valid
}
```

wcwy: BTW, you might want to consider looking at the built-in startsWith() method.

src/main/java/Duke.java
```diff
@@ -1,10 +1,260 @@
...
+    //             ui.showError(e.getMessage());
+    //         } finally {
+    //             ui.showLine();
+    //         }
+    //     }
+    // }
+
+    // public static void main(String[] args) {
+    //     new Duke(FILE_PATH).run();
+    // }
```
wcwy: Don't forget to remove your unused code!

src/main/java/Duke.java
```diff
@@ -1,10 +1,260 @@
...
+
     public static void main(String[] args) {
-        String logo = " ____        _        \n"
-                + "|  _ \\ _   _| | _____ \n"
-                + "| | | | | | | |/ / _ \\\n"
-                + "| |_| | |_| |   <  __/\n"
-                + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
+
+        //read the file
```
wcwy: For comments, the first letter should capped and there should be a space between the `//` and the words after it.

src/main/java/Parser.java
```diff
@@ -0,0 +1,3 @@
+public class Parser {
```
wcwy, line 9: Might want to migrate your code from `Check` class to this. The naming for the `Check` class violates the coding standard and parser is a much better name.

src/main/java/Duke.java
```diff
@@ -1,10 +1,260 @@
+import java.util.Scanner;  // Import the Scanner class
```
wcwy: There are some areas of improvement on this file: 
1. Everything is squeezed into one single main methods. That's too long. Try to refractor them into multiple methods, or even into other classes.
2. Arrowhead code. Generally we try not to have more than 3 levels of indentation in our code. Again, refactoring may help in this.

src/main/java/Duke.java
```diff
@@ -1,10 +1,260 @@
...
+                        throw new IllegalArgumentException("The task number is less than 1.");
+                    }
+                    //if there is no task at the task number throw an exception
+                    if(tasks.get(taskNumber-1)==null){
+                        throw new IllegalArgumentException("There is no task at the task number.");
+                    }
+                    //if the task is already not done throw an exception
+                    if(tasks.get(taskNumber-1).isDone==false){
+                        throw new IllegalArgumentException("The task is already not done.");
+                    }
```
wcwy: For all your error messages, you might want to consider refactoring them into another class/enum.

src/main/java/DukeException.java
```diff
@@ -0,0 +1,36 @@
+//Implement a class DukeException that extends the Exception class. 
+//This class should be able to handle the different exceptions that may occur in your program.
+
+public class DukeException extends Exception {
+    //CommandException
+    // Constructor that accepts a message
+    public DukeException(String message)
+    {
+       super(message);
+    }
```
wcwy: The line breaks around curly brackets for constructors follow the same rule as method. It should be like this:
```
public DukeException(String message) {
    super();
}
```

src/main/java/Storage.java
```diff
@@ -0,0 +1,144 @@
+import java.util.Scanner;  // Import the Scanner class
+import java.io.File; // Import the File class
+import java.io.FileWriter; // Import the FileWriter class
+import java.io.BufferedWriter; // Import the BufferedWriter class
+import java.io.IOException; // Import the IOException class to handle errors
+import java.util.ArrayList; // Import the ArrayList class
```
wcwy, line 18: Comments are generally not needed for imports

src/main/java/Storage.java
```diff
@@ -0,0 +1,144 @@
...
+        //try to read the file
+        try{
+            //Create a scanner
+            Scanner fileScan = new Scanner(file);
+            //Read the file
+            while(fileScan.hasNextLine()){
+                //Read the line
+                String line = fileScan.nextLine();
+                //Split the line
+                String[] lineSplit = line.split(" \\| ");
```
wcwy, line 75: Magic string here. Consider making it into a named constants.

src/main/java/Storage.java
```diff
@@ -0,0 +1,144 @@
...
+                //if the task is a todo
+                if(lineSplit[0].equals("T")){
```
wcwy, line 77: Magic string here too.

There are also other magic strings/numbers/literals in this file and other files. Try to use named constants or enum (if applicable) for them.

src/main/java/Storage.java
```diff
@@ -0,0 +1,144 @@
...
+                file.getParentFile().mkdirs();
+                file.createNewFile();
+            }catch(Exception e1){
+                e1.printStackTrace();
+            }
+        }
+    }
+   
+    //method that tand reads the file and returns an arraylist of tasks 
+    public static ArrayList<Task> readFile(String filePath){
```
wcwy, line 59: This method seems too long. Consider refactoring it into multiple methods. Also consider applying SLAP when you are refactoring this.

src/main/java/Storage.java
```diff
@@ -0,0 +1,144 @@
...
+                    //write the task to the file
+                    bw.write("D | " + tasks.get(i).isDone + " | " + tasks.get(i).taskName + " | " + ((Deadline) tasks.get(i)).deadline);
+                    bw.newLine();
+                }
+                //if the task is an event
+                else if(tasks.get(i) instanceof Event){
+                    //write the task to the file
+                    bw.write("E | " + tasks.get(i).isDone + " | " + tasks.get(i).taskName + " | " + ((Event) tasks.get(i)).eventTime);
+                    bw.newLine();
+                }
```
wcwy, line 157: Each type of Task should know how to print itself, for example, it should have a `toString()` method that `writeFile()` calls, rather than this method itself having code for each task type -- this isn't extensible; if you add a new task type, you would need to change this method. Exploit Java's polymorphism in cases like this, it can make your code neater and your program more scalable.

src/main/java/Task.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+    //toString method to print the status of the task followed by the task name
+    public String toString(){
+        if(this.isDone){
+            return "[E][X] " + this.taskName + " (" + this.eventTime + ")";
+        }else{
+            return "[E][ ] " + this.taskName + " (" + this.eventTime + ")";
+        }
+    }
+}
```
wcwy: Instead of putting them inside this `Task.java` file, you should create a `Todo.java` file and something similar for deadline and event. Each classes should have a file on its own.

wcwy (commented):
Good progress there! I notice frequent spacing issues within your code (more info below). I know it can be annoying that there is a standard to follow on this, but oh wells, please try your best to follow it. More comments below.

PR #94, opened 2023-01-29
Add increment

tsx0314:


src/main/java/Duke.java
```diff
@@ -1,10 +1,84 @@
...
+		dukeGreeting.printGreetings();
+		dukeGreeting.printOpeningLine();
+		Scanner sc = new Scanner(System.in);
+
+		Task[] tasks = new Task[100];
+		int taskCount = 0;
+
+		String userMessage;
+
+		boolean shouldExit = false;
```
Manoj364: I like how the method's name is a verb and written in camelCase

src/main/java/Duke.java
```diff
@@ -1,10 +1,84 @@
...
+					for (int i = 0; i < taskCount; i += 1) {
+						int serialNumber = i + 1;
+						System.out.print(serialNumber + ".");
+						tasks[i].printTask();
+					}
+				}
+				dukeGreeting.printDivider();
+			} else if (userMessage.startsWith("mark")) {
+				int no = findTaskNumber(userMessage);
+				if (no == 0 || no < 0) {
```
Manoj364: This can be expressed more explicitly

src/main/java/Duke.java
```diff
@@ -1,10 +1,84 @@
...
+				if (taskCount == 0 || taskCount < 0) {
+					System.out.println("No task！");
+				} else {
+					System.out.println("Here are the tasks in your list: ");
+					for (int i = 0; i < taskCount; i += 1) {
+						int serialNumber = i + 1;
+						System.out.print(serialNumber + ".");
+						tasks[i].printTask();
+					}
+				}
```
Manoj364: Perhaps you can avoid deep-nesting

src/main/java/Greetings.java
```diff
@@ -0,0 +1,48 @@
...
+					+ "|____/ \\__,_|_|\\_\\___|\n";
+	private static final String OPENING_LINE =
+			"Hello! I'm Duke"
+					+System.lineSeparator()
+					+ "What can I do for you?";
+	private static final String EXIT_LINE =
+			"Bye. Hope to see you again soon!";
+	private static final String DIVIDER =
+			"______________________________";
+	private static final String ERROR = "Hey! Error!";
```
Manoj364: I like how you declared the constants appropriately

src/main/java/Duke.java
```diff
@@ -1,10 +1,84 @@
...
+		boolean shouldExit = false;
+
+		while (!shouldExit) {
+			userMessage = sc.nextLine();
+
+			//check the user message
+			if (userMessage.equals("bye")) {
+				shouldExit = true;
+				dukeGreeting.printExitLine();
+			} else if (userMessage.equals("hello") || userMessage.equals("echo")) {
```
Manoj364: I feel that this can be expressed explicitly

src/main/java/Duke.java
WilsonOh: Good naming for boolean value, shows purpose of the variable clearly.

src/main/java/Greetings.java
WilsonOh: Good use of constants.

src/main/java/Greetings.java
```diff
@@ -0,0 +1,48 @@
...
+		System.lineSeparator();
+		System.out.println(DIVIDER);
+	}
+
+	public void printErrorMessage() {
+		printDivider();
+		System.out.println(ERROR);
+		printDivider();
+	}
+}
```
WilsonOh: There's some repetition here i.e. printDivider() -> sysout something -> printDivider(). Might be good to extract this pattern out to a method.

src/main/java/Task.java
```diff
@@ -0,0 +1,37 @@
...
+	public void unmarkAsUndone() {
+		isCompleted = false;
+		System.out.println("OK, I've marked this task as not done yet: ");
+	}
+
+	void printTask() {
+		System.out.println("["
+				+getStatusIcon() + "] "
+				+ getDescription());
+	}
```
WilsonOh: Clear and readable method names which does exactly what the name suggests.

src/main/java/duke/tasks/Event.java
```diff
@@ -0,0 +1,40 @@
...
+
+	public void setStartTime(String startTime) {
+		this.from = startTime;
+	}
+
+	public void setEndTime(String endTime) {
+		this.to = endTime;
+	}
+
+	// Prints task
```
bdthanh: You should remove this comment

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,126 @@
...
+import duke.tasks.Event;
+import duke.tasks.Task;
+import duke.tasks.Todo;
+import duke.ui.Greetings;
+
+import java.util.Scanner;
+
+public class Duke {
+
+	public static void startDuke() {
```
bdthanh: This method is a bit long and has a lot of responsibilities. You might consider create a new methods for each case. For example in case userMessage.startsWith("todo"), you can make a method to validate the command and call it in the case block. So that the startDuke method will be shorter and prettier

src/main/java/duke/ui/Greetings.java
```diff
@@ -0,0 +1,66 @@
+package duke.ui;
+
+public class Greetings {
```
bdthanh: Class name seems to be off since you have a lot than Greeting. You might consider to change the name

bdthanh (commented):
Looks good to me! Be careful with class naming and in line comment. And you can consider to shorten your method to improve your code quality

PR #118, opened 2023-01-30
[cleoncheng2000] iP

cleoncheng2000:


src/main/java/Duke.java
```diff
@@ -1,10 +1,115 @@
...
+    private static void printAllTasks(Task[] tasks) {
+        System.out.println("____________________________________________________________");
+        System.out.println("Here are the tasks in your list:");
+        for (Task task : tasks) {
+            if (task != null) {
+                System.out.println(task);
+            }
+        }
+        System.out.println("____________________________________________________________");
+    }
```
choongzhanhong: I really like that you have extracted all these methods out! However, could it be better to also extract the method to print a horizontal line?

src/main/java/Duke.java
```diff
@@ -1,10 +1,115 @@
...
+            if (input.equals("bye")) {
+                printByeStatement();
+                break;
+            } else if (input.equals("list")) {
+                printAllTasks(tasks);
+            } else if (input.startsWith("mark")) {
+                String[] temp = input.split(" ", 2);
+                int taskIndex = Integer.parseInt(temp[1]);
+                Task curTask = tasks[taskIndex - 1];
+                curTask.markAsDone();
```
choongzhanhong: Would it be good to also extract these as a method? Likewise, for the below blocks too.

src/main/java/Duke.java
```diff
@@ -1,10 +1,115 @@
...
+            else if (input.startsWith("deadline")){
+                String[] temp1 = input.split(" /by ", 2);
+                String by = temp1[1];
+                String[] temp2 = temp1[0].split(" ", 2);
+                String description = temp2[1];
+                Deadline deadline = new Deadline(currentIndex + 1, description, by);
+                tasks[currentIndex] = deadline;
+                currentIndex++;
+                printTaskAddedStatement(currentIndex, deadline);
+            }
```
choongzhanhong: Perhaps it might be useful to indicate the intended format input via some comments?
So that we can understand why you are splitting through `"/by"` and what the outcome is supposed to be.

src/main/java/Duke.java
```diff
@@ -1,10 +1,115 @@
...
+                String[] temp1 = input.split(" /to ", 2);
+                String to = temp1[1];
+                String[] temp2 = temp1[0].split(" /from ", 2);
+                String from = temp2[1];
+                String[] temp3 = temp2[0].split(" ", 2);
+                String description = temp3[1];
+                Event event = new Event(currentIndex + 1, description, from, to);
+                tasks[currentIndex] = event;
+                currentIndex++;
+                printTaskAddedStatement(currentIndex, event);
```
choongzhanhong: Could `temp` as a variable name be insufficiently descriptive?
Likewise, the numbering of the variables as well.

src/main/java/Duke.java
```diff
@@ -1,10 +1,115 @@
...
+        System.out.println(" Hello! I'm Duke");
+        System.out.println(" What can I do for you?");
+        System.out.println("____________________________________________________________");
+    }
+
+    private static void printByeStatement() {
+        System.out.println("____________________________________________________________");
+        System.out.println(" Bye. Hope to see you again soon!");
+        System.out.println("____________________________________________________________");
+    }
```
choongzhanhong: I like how your similar methods are named consistently.

choongzhanhong (commented):
Overall, very well-organised.
A few comments on methods that could be extracted, as well as improving code readability for other users.

src/main/java/Duke.java
```diff
@@ -1,10 +1,221 @@
...
-    public static void main(String[] args) {
-        String logo = " ____        _        \n"
-                + "|  _ \\ _   _| | _____ \n"
-                + "| | | | | | | |/ / _ \\\n"
-                + "| |_| | |_| |   <  __/\n"
-                + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
-    }
-}
+    public static void main(String[] args) throws IOException {
```
okkhoy, line 21: This is a very long method. 
Please extract out relevant code to separate methods.

src/main/java/Duke.java
```diff
@@ -1,10 +1,221 @@
...
+        Scanner in = new Scanner(System.in);
+        int tasksSize = 100;
```
okkhoy: Constant?
In fact, since you are using arraylist, this may not be relevant, is it?

src/main/java/Duke.java
```diff
@@ -1,10 +1,221 @@
...
+            } else if ((input.startsWith("todo") | input.startsWith("deadline") | input.startsWith("event"))
+                    & (tasks.size() > maxIndex)) {
+                printDottedLine();
+                System.out.println("☹ OOPS!!! The tasks list is currently full");
+                printDottedLine();
+            } else {
+                printUnknownCommandError();
+            }
+        }
+    }
```
okkhoy: There is some deeply nested code in the method, and ample opportunities to improve SLAP.
Please consider refactoring the code.

src/main/java/Save.java
```diff
@@ -0,0 +1,81 @@
...
+                    if (args[2].equals("X")) {
+                        event.markAsDone();
+                    }
+                    tasks.add(event);
+                }
+            }
+        } catch (IOException exception) {
+            System.out.println("☹ OOPS!!! An error occurred");
+        } catch (Exception exception) {
+            System.out.println("☹ OOPS!!! An error occurred 2");
```
okkhoy: This is not a good exception message. Please reword it.

src/main/java/Save.java
```diff
@@ -0,0 +1,81 @@
...
+                    tasks.add(deadline);
+                } else {
+                    Event event = new Event(args[2], args[3], args[4]);
+                    if (args[2].equals("X")) {
+                        event.markAsDone();
+                    }
+                    tasks.add(event);
+                }
+            }
+        } catch (IOException exception) {
```
okkhoy: the catch only belongs to the file read. once the file is read, rest of the operation can go ahead without troubles.
reordering the `try ... catch`  will eliminate the arrowhead code.

okkhoy (commented):
1. Please add Javadoc comments to al non-trivial methods and classes in the codebase
2. I don't see any tests; please add reasonable number of tests

PR #119, opened 2023-01-30
[yixuann02] iP

yixuann02:


src/main/java/Duke.java
```diff
@@ -1,10 +1,66 @@
+import java.util.Scanner;
+
 public class Duke {
+
+    public static final String HORIZONTAL_LINES = "____________________________________________________________\n";
+    public static final String INDENTATIONS = "    ";
```
joulev: I think `HORIZONTAL_LINE` and `INDENTATION` might be better names for these two variables.

src/main/java/Duke.java
```diff
@@ -1,10 +1,66 @@
...
-        String logo = " ____        _        \n"
-                + "|  _ \\ _   _| | _____ \n"
-                + "| | | | | | | |/ / _ \\\n"
-                + "| |_| | |_| |   <  __/\n"
-                + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
+        System.out.println(INDENTATIONS + HORIZONTAL_LINES +
+                INDENTATIONS + "Hello! I'm Duke\n" +
+                INDENTATIONS + "What can I do for you?\n" +
+                INDENTATIONS + HORIZONTAL_LINES);
```
joulev: I believe a helper function like

```java
static void printlnWithIndentation(String string) {
    System.out.print(INDENTATIONS);
    System.out.println(string);
}
```

would be helpful in this `main` function to avoid code repetition, as I see a lot of `System.out.println(INDENTATIONS + "something")` in this file.

src/main/java/Duke.java
```diff
@@ -1,10 +1,66 @@
...
+        Scanner in = new Scanner(System.in);
+        String readLine;
+        Task[] tasks = new Task[100];
+        int taskIndex = 0;
+        while(true) {
```
joulev:
```suggestion
        while (true) {
```

I noticed similar cases in many other lines as well, it might be good to modify them to fully follow the style guide.

src/main/java/Duke.java
```diff
@@ -1,10 +1,66 @@
...
+                        INDENTATIONS + HORIZONTAL_LINES);
+                
+            }else {
+                // add new tasks into todo
+                tasks[taskIndex] = new Task(readLine);
+                System.out.println(INDENTATIONS + HORIZONTAL_LINES +
+                        INDENTATIONS + "added: " + readLine + '\n' +
+                        INDENTATIONS + HORIZONTAL_LINES);
+                taskIndex++;
+            }
```
joulev: A `switch`/`case` statement might be more readable than `if`/`else` in this situation in my opinion.

src/main/java/Duke.java
```diff
@@ -1,10 +1,66 @@
...
-                + "| | | | | | | |/ / _ \\\n"
-                + "| |_| | |_| |   <  __/\n"
-                + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
+        System.out.println(INDENTATIONS + HORIZONTAL_LINES +
+                INDENTATIONS + "Hello! I'm Duke\n" +
+                INDENTATIONS + "What can I do for you?\n" +
+                INDENTATIONS + HORIZONTAL_LINES);
+        Scanner in = new Scanner(System.in);
+        String readLine;
```
joulev: Perhaps `userInput` would be a better name here, as I would expect `readLine` to be the name of a function that reads user input instead.

joulev (commented):
Very nice code! Here are some of my thoughts regarding it in more details.

src/main/java/Duke.java
```diff
@@ -1,10 +1,66 @@
...
+        Task[] tasks = new Task[100];
```
itszhixuan: Using a named constant could possibly help avoid magic numbers and improve code quality.

src/main/java/Duke.java
itszhixuan: The while loop is structured logically and is easy to read. Good job!

src/main/java/Duke.java
```diff
@@ -1,10 +1,66 @@
+import java.util.Scanner;
+
 public class Duke {
+
+    public static final String HORIZONTAL_LINES = "____________________________________________________________\n";
+    public static final String INDENTATIONS = "    ";
+
     public static void main(String[] args) {
```
itszhixuan: You may consider using SLAP to make main easier to read.

src/main/java/Duke.java
```diff
@@ -1,10 +1,66 @@
+import java.util.Scanner;
+
 public class Duke {
+
+    public static final String HORIZONTAL_LINES = "____________________________________________________________\n";
```
itszhixuan: Good job in using constants to assign the fixed formatting code

src/main/java/Duke.java
```diff
@@ -1,10 +1,66 @@
...
+                System.out.println(INDENTATIONS + HORIZONTAL_LINES);
+
+            } else if (readLine.startsWith("mark")) {
+                String toMark = readLine.substring(5);
+                toMark = toMark.trim();
+                int indexToMark = Integer.parseInt(toMark) - 1;
+                tasks[indexToMark].setDone(true);
+                System.out.println(INDENTATIONS + "Nice! I've marked this task as done:\n" +
+                        INDENTATIONS + "  [X] " + tasks[indexToMark].getTaskDescription() + '\n' +
+                        INDENTATIONS + HORIZONTAL_LINES);
```
itszhixuan: I like how you ensured that the lines to be printed out are wrapped when appropriate, making the code easier to read.

itszhixuan (commented):
Overall well done!

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,215 @@
...
+                    break;
+                case "todo":
+                    // command todo
+                    taskIndex = commandTodo(tasks, taskIndex, inputLine);
+                    break;
+                case "deadline":
+                    taskIndex = commandDeadline(tasks, taskIndex, inputLine);
+                    break;
+                case "event":
+                    taskIndex = commandEvent(tasks, taskIndex, inputLine);
```
kerct: Perhaps you can rename these functions to state what they do; using verbs

src/main/java/duke/exception/IllegalCommandException.java
```diff
@@ -0,0 +1,7 @@
+package duke.exception;
+
+public class IllegalCommandException extends Exception{
```
kerct, line 3:
```suggestion
public class IllegalCommandException extends Exception {
```

src/main/java/duke/exception/IllegalCommandException.java
```diff
@@ -0,0 +1,7 @@
...
+    public IllegalCommandException(String str){
```
kerct, line 4:
```suggestion
    public IllegalCommandException(String str) {
```

src/main/java/duke/command/Todo.java
```diff
@@ -0,0 +1,16 @@
+package duke.command;
+
+import duke.task.Task;
+
+public class Todo extends Task {
+    protected String add;
```
kerct, line 6: Unused variables should be removed

kerct (commented):
Code is pretty neat and readable overall 👍 Perhaps you can consider moving some code out of your Duke.java into another class as it is quite long now

PR #126, opened 2023-01-30
[liuziyang020319] iP

liuziyang020319:
a Level 3 iP, Pull Request.

### DIFF
--- a/src/main/java/Duke.java
+++ b/src/main/java/Duke.java
@@ -6,5 +6,9 @@ public class Duke {
                 + "| |_| | |_| |   <  __/\n"
                 + "|____/ \\__,_|_|\\_\\___|\n";
         System.out.println("Hello from\n" + logo);
+
+        System.out.println("Hello! I'm Duke\n");
+        System.out.println("What can I do for you?\n");
+        System.out.println("Bye. Hope to see you again soon!\n");
     }
 }
--- a/src/main/java/Duke.java
+++ b/src/main/java/Duke.java
@@ -1,3 +1,4 @@
+import java.util.Scanner;
 public class Duke {
     public static void main(String[] args) {
         String logo = " ____        _        \n"
@@ -9,6 +10,25 @@ public class Duke {
 
         System.out.println("Hello! I'm Duke\n");
         System.out.println("What can I do for you?\n");
-        System.out.println("Bye. Hope to see you again soon!\n");
+
+        Print.PrintLine();
+
+        Scanner in = new Scanner(System.in);
+        while(true){
+            String line = in.nextLine();
+            line = line.trim();
+            String[] command = line.split(" ");
+            switch (command[0]){
+            case "bye":
+                Print.PrintString("Bye. Hope to see you again soon!");
+                return;
+            case "list":
+                Print.PrintString("list");
+                break;
+            default:
+                Print.PrintString(line);//simply echos commands entered by the user
+            }
+        }
+
     }
 }
--- a/src/main/java/Duke.java
+++ b/src/main/java/Duke.java
@@ -3,16 +3,17 @@ public class Duke {
     public static void main(String[] args) {
         DukePrinter.printGreeting();
         Scanner in = new Scanner(System.in);
-        while(true){
+        while (true) {
+            //todo not follow SLAP here, fixed it later.
             String line = in.nextLine();
             line = line.trim();
             String[] command = line.split(" ");
-            String lineRemainder = line.substring(line.indexOf(" ")+1);
-            lineRemainder = lineRemainder.trim();
+            String lineRemaining = line.substring(line.indexOf(" ")+1);
+            lineRemaining = lineRemaining.trim();
             int id;
             switch (command[0]){
             case "bye":
-                DukePrinter.printString("Bye. Hope to see you again soon!");
+                DukePrinter.printStringln("Bye. Hope to see you again soon!");
                 DukePrinter.printGoodbyeLogo();
                 return;
             case "list":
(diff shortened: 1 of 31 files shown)


src/main/java/Duke.java
```diff
@@ -1,10 +1,83 @@
...
-                + "| |_| | |_| |   <  __/\n"
-                + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
+        DukePrinter.printGreeting();
+        Scanner in = new Scanner(System.in);
+        while (true) {
+            //todo not follow SLAP here, fixed it later.
+            String line = in.nextLine();
+            line = line.trim();
+            String[] command = line.split(" ");
```
iantenkai: I like the use of a string to store the initial command

iantenkai: Would issues arise if the user input is longer than the maximum size of the array?

src/main/java/Duke.java
```diff
@@ -1,10 +1,83 @@
...
+                    id = Integer.parseInt(lineRemaining);
+                } catch (NumberFormatException integerException) {
+                      DukePrinter.printErrorln("Sorry, the id is invalid!");
+                    break;
+                }
+                DukeList.markDone(id-1);
+                break;
+            case "unmark":
+                try {
+                    id = Integer.parseInt(lineRemaining);
```
iantenkai: Since you have already parsed the input into a string array, wouldn't it be better to use the second index of the  command array instead?

src/main/java/Duke.java
```diff
@@ -1,10 +1,83 @@
...
+                    DukePrinter.printErrorln("Please use the format: " +
+                            "deadline <task name> /by <deadline time>");
+                }
+                break;
+            case "event":
+                try {
+                    //todo not follow SLAP here, fixed it later.
+                    String eventName = lineRemaining.substring(0, lineRemaining.indexOf("/from")-1);
+                    eventName = eventName.trim();
+                    String eventTimeFrom = lineRemaining.substring(lineRemaining.indexOf("/from")+6,
```
iantenkai: could these statements be better abstracted?

src/main/java/Duke.java
```diff
@@ -1,10 +1,83 @@
...
+                DukeList.listTask();
+                break;
+            case "mark":
+                try {
+                    id = Integer.parseInt(lineRemaining);
+                } catch (NumberFormatException integerException) {
+                      DukePrinter.printErrorln("Sorry, the id is invalid!");
+                    break;
+                }
+                DukeList.markDone(id-1);
```
iantenkai: the mark and unmark cases are the same. I think it would be better to refactor them to a single method

src/main/java/Duke.java
liuziyang020319: > could these statements be better abstracted?

Yep, I already find this problem, and I'll fix it later. Thank you! 👍

src/main/java/Duke.java
liuziyang020319: > Would issues arise if the user input is longer than the maximum size of the array?

Good question, I'll check whether this exception will happen or not.

src/main/java/Duke.java
```diff
@@ -1,10 +1,83 @@
...
+        Scanner in = new Scanner(System.in);
+        while (true) {
+            //todo not follow SLAP here, fixed it later.
+            String line = in.nextLine();
+            line = line.trim();
+            String[] command = line.split(" ");
+            String lineRemaining = line.substring(line.indexOf(" ")+1);
+            lineRemaining = lineRemaining.trim();
+            int id;
+            switch (command[0]){
```
matthew-liu-zhenjie: I really like the use of an array of strings to implement a switch case to process the user's command. It looks great!

src/main/java/DukeList.java
```diff
@@ -0,0 +1,48 @@
...
+    public static boolean isValidID(int id) {
+        boolean isIDInValid = id < 0 || id >= taskList.size();
+        if(isIDInValid) {
+            DukePrinter.printErrorln("Sorry, the id is invalid!");
+            return false;
+        }
+        return true;
+    }
+
+    public static void markDone(int id) {
```
matthew-liu-zhenjie: Nice use of SLAP. Good job!

src/main/java/Duke.java
matthew-liu-zhenjie: Perhaps (id-1) could be put inside the method? Could help with readability.

src/main/java/Duke.java
```diff
@@ -1,10 +1,83 @@
...
+                if(lineRemaining.equals("")) {
+                    DukePrinter.printErrorln("Sorry, the description of a todo cannot be empty.");
+                    break;
+                }
+                DukeList.addTask(new DukeTask(lineRemaining));
+                break;
+            case "deadline":
+                try {
+                    //todo not follow SLAP here, fixed it later.
+                    String deadlineName = lineRemaining.substring(0, lineRemaining.indexOf("/by"));
```
matthew-liu-zhenjie: Perhaps you could refactor this into a single method

src/main/java/DukePrinter.java
```diff
@@ -0,0 +1,47 @@
+public class DukePrinter {
```
matthew-liu-zhenjie: Good use of a class to group the printing outputs

src/main/java/Duke/DukeRobot.java
```diff
@@ -0,0 +1,133 @@
...
+import Duke.DukeCommandLine.DukeTaskInputException;
+import Duke.DukeFunction.DukeList;
+import Duke.DukeFunction.DukePrinter;
+import Duke.DukeTask.DukeDeadline;
+import Duke.DukeTask.DukeEvent;
+import Duke.DukeTask.DukeTask;
+
+import java.util.Scanner;
+
+public class DukeRobot {
```
jeyvia, line 10: I notice that you have a tendency to use magic strings and numbers. Do try to avoid that where possible, and instead use constants to set your strings/numbers.

src/main/java/Duke/DukeRobot.java
```diff
@@ -0,0 +1,133 @@
...
+                break;
+            case "deadline":
+                try {
+                    DukeDeadline deadline = processDeadline(command.getCommandMessage());
+                    DukeList.addTask(deadline);
+                } catch (DukeTaskInputException e) {
+                    DukePrinter.printErrorln(e.getMessage());
+                }
+                break;
+            case "event":
```
jeyvia: Good job with the switch-case for your commands!

src/main/java/Duke/DukeFunction/DukeList.java
```diff
@@ -0,0 +1,137 @@
...
+import java.io.File;
+import java.io.IOException;
+import java.io.FileWriter;
+import java.util.ArrayList;
+import java.util.Scanner;
+
+public class DukeList {
+    private static final ArrayList<DukeTask> taskList = new ArrayList<>();
+    private static final String FILE_PATH = "data/list.txt";
+    private static final int MAX_TASK = 100;
```
jeyvia: In the project constraints, there is no indication of a maximum size. Please do avoid limiting your `taskList` to a size if possible :)

src/main/java/Duke/DukeFunction/DukePrinter.java
```diff
@@ -0,0 +1,49 @@
+package Duke.DukeFunction;
+
+public class DukePrinter {
```
jeyvia: I like that you have reorganised the things to print into a different `.java`. However, it would be better to standardise, and have all your print statements in this `.java`. Currently, I do still spot a few print statements in other files.

src/main/java/Duke/DukeRobot.java
```diff
@@ -0,0 +1,133 @@
...
+                    inputTask.indexOf("/to")-1);
+            eventTimeFrom = eventTimeFrom.trim();
+            String eventTimeTo = inputTask.substring(inputTask.indexOf("/to")+4);
+            return new DukeEvent(eventName, eventTimeFrom, eventTimeTo);
+        } catch (StringIndexOutOfBoundsException e) {
+            throw new DukeTaskInputException("Please use the format: " +
+                    "event <task name> /from <event time from> /to <event time to>");
+        }
+    }
+    public static void main(String[] args) {
```
jeyvia, line 58: Your code is very well organised. You process your inputs in a standardised manner and catch exceptions where necessary. Well done! Do consider adding more exceptions in the future increments :)

PR #137, opened 2023-01-30
[Aung Phone Naing] iP

Aung-Phone-Naing:
Duke able to greet and echo

### DIFF
--- a/src/main/java/Duke.java
+++ b/src/main/java/Duke.java
@@ -4,6 +4,7 @@ import Tasks.Event;
 import Tasks.Task;
 import Tasks.ToDo;
 
+import java.util.ArrayList;
 import java.util.Scanner;
 
 public class Duke {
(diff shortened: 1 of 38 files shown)


src/main/java/Duke.java
```diff
@@ -1,10 +1,62 @@
...
+            if (line.equals("list")) {
+                if (!list.isEmpty()) {
+                    System.out.println();
+                    printList(list);
+                } else {
+                    System.out.println('\n' + "List is empty!" + '\n');
+                }
+            } else if (line.startsWith("mark")) {
+                int list_index = Integer.parseInt(line.substring(5)) - 1;
+                if (list.get(list_index).getIsDone()) {
```
TopGun2001: I like how you have added an extra if statement to check if the task is marked/unmarked. This reduces the task for your program to mark/unmark it again which is quite efficient.

src/main/java/Duke.java
```diff
@@ -1,10 +1,62 @@
...
+                } else {
+                    list.get(list_index).markAsUndone();
+                    System.out.println('\n' + "OK, I've marked this task as not done yet:");
+                    System.out.println("  [" + list.get(list_index).getStatusIcon() + "] " + list.get(list_index).getDescription() + '\n');
+                }
+            } else {
+                System.out.println('\n' + "added: " + line + " to the task list!" + '\n');
+                Task temp_task = new Task(line);
+                list.add(temp_task);
+            }
```
TopGun2001: You have nested the **if-else** statements and they look a bit unstructured. I'd suggest you to use switch statements if that works out for you.

src/main/java/Duke.java
```diff
@@ -1,10 +1,62 @@
...
+        while(!line.equals("bye")) {
+            if (line.equals("list")) {
+                if (!list.isEmpty()) {
+                    System.out.println();
+                    printList(list);
+                } else {
+                    System.out.println('\n' + "List is empty!" + '\n');
+                }
+            } else if (line.startsWith("mark")) {
+                int list_index = Integer.parseInt(line.substring(5)) - 1;
```
Zemdalk: Perhaps using camelCase for variable `list_index` is better?

src/main/java/Duke.java
```diff
@@ -1,10 +1,62 @@
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Scanner;
+
 public class Duke {
+    public static void printList(List<Task> l1) {
```
Zemdalk: I like the way you use `ArrayList` to manage tasks.

src/main/java/Duke.java
Zemdalk: I argee with @TopGun2001 basically. And maybe it's better to extract your codes inside these `if-else` statements and form several functions apart from `main` function? This would make `main` function not too long maybe.

src/main/java/Duke.java
Zemdalk: Also I really like the way you use function `printList` to print the list.

Zemdalk (commented):
Looks good. I just give some personal suggestions on naming and functions

src/main/java/Duke.java
```diff
@@ -1,10 +1,119 @@
...
-                + "| | | | | | | |/ / _ \\\n"
-                + "| |_| | |_| |   <  __/\n"
-                + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
+        /*String logo = " ____        _        \n"
+                      + "|  _ \\ _   _| | _____ \n"
+                      + "| | | | | | | |/ / _ \\\n"
+                      + "| |_| | |_| |   <  __/\n"
+                      + "|____/ \\__,_|_|\\_\\___|\n";
+        System.out.println("Hello from\n" + logo);*/
```
jeyvia: Instead of leaving it commented out, you can remove it :)

src/main/java/Duke.java
```diff
@@ -1,10 +1,119 @@
...
+        Scanner in = new Scanner(System.in);
+        String line, commandWord, description;
+        int taskListIndex, currTaskNumber;
+        String[] lineSpaced;
+        boolean isChanged = false;
+
+        line = in.nextLine();
+        lineSpaced = parseCommands(line);
+        currTaskNumber = 0;
+        while(!lineSpaced[0].equals("bye")) {
```
jeyvia: You should not default your command behaviours like this! Create a case for every command :)

src/main/java/Task.java
```diff
@@ -0,0 +1,38 @@
...
+    public String getDescription() {
+        return description;
+    }
+
+    public boolean getIsDone() {
+        return isDone;
+    }
+
+
+    // class initialization
```
jeyvia: Please remove such comments as they go against the java coding standards.

src/main/java/Duke.java
```diff
@@ -1,10 +1,119 @@
+import java.lang.reflect.Array;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Scanner;
+
 public class Duke {
+    public static void printList(Task[] l1, int currListIndex) {
+        int index;
+        if (l1[0] == null) {
```
jeyvia: I notice that you have a tendency to use magic strings and numbers. Do try to avoid that where possible, and instead use constants to set your strings/numbers.

src/main/java/Duke.java
```diff
@@ -1,10 +1,119 @@
...
+                    isChanged = true;
+                }
+                printUnmarked(isChanged, tasksList[taskListIndex].getStatusIcon(), tasksList[taskListIndex].getDescription());
+                break;
+            case "Todo":
+                description = lineSpaced[1];
+                tasksList[currTaskNumber] = new ToDo (description);
+                printAdded(tasksList[currTaskNumber], currTaskNumber+1);
+                currTaskNumber++;
+                break;
```
jeyvia: I do not see a command case for `Event` or `Deadline`. Could it be missing from this commit? If so, it's fine you can add it in later versions. 👍

src/main/java/Duke.java
Aung-Phone-Naing: Okie! I will go and create a constants file for messages and numbers to fix this! Thank you!

src/main/java/Task.java
Aung-Phone-Naing: Okie I shall use java docs to do it Thank you!

PR #145, opened 2023-01-30
[YongbinWang] iP

YongbinWang:


src/main/java/Alex.java
```diff
@@ -0,0 +1,79 @@
...
+        Scanner myObj = new Scanner(System.in);  // Create a Scanner object
+        greeting();
+        boolean isBye = false;
+        while(!isBye){
+            String userInput = myObj.nextLine();
+            String[] words = userInput.split(" ");
+            if(words[0].toLowerCase().equals("list")){
+                for(int i = 1; i <= numberOfTasks; i++){
+                    System.out.print(i);
+                    System.out.println(". " + "[" + allTasks[i-1].getStatusIcon() + "]" + " " + allTasks[i-1].getDescription());
```
EthanYidong: This code appears in very similar forms below. Perhaps it would be a good idea to extract this out to be a method of `Task`?

src/main/java/Alex.java
```diff
@@ -0,0 +1,79 @@
...
+        String horizontalLine = "____________________________________________________________";
+        System.out.println(horizontalLine);
+    }
+    public static void greeting(){
+        printLine();
+        String greet = "Hello! I'm Alex\nWhat can I do for you today?";
+        //String question = "What can I do for you today?";
+        System.out.println(greet);
+        //System.out.println(question);
+        printLine();
```
EthanYidong: I think keeping old code present as comments decreases readability as most readers are expecting explanations of code in your comments. Maybe you can delete these and refer to previous commits if you ever need to reference old versions of your code?

src/main/java/Alex.java
```diff
@@ -0,0 +1,79 @@
+import java.util.Scanner;
+
+public class Alex {
+    public static void main(String[] args) {
+        Task[] allTasks = new Task[100];
```
EthanYidong: I don't like the usage of a magic literal here, perhaps you can extract it to a variable which better explains what the 100 means here?

src/main/java/Alex.java
```diff
@@ -0,0 +1,79 @@
...
+                    System.out.print(i);
+                    System.out.println(". " + "[" + allTasks[i-1].getStatusIcon() + "]" + " " + allTasks[i-1].getDescription());
+                }
+                printLine();
+
+            }
+            else if(words[0].toLowerCase().equals("bye")){
+                exit();
+                isBye = true;
+                break;
```
EthanYidong: `isBye = true` is redundant here because `break` will exit from the loop. It may be best if you remove `isBye` from your code as it isn't serving any purpose here.

src/main/java/Alex.java
```diff
@@ -0,0 +1,79 @@
...
+        }
+
+
+    }
+    public static void echoResponse(String userInput){
+        System.out.println("added: " + userInput);
+        printLine();
+    }
+    public static void printLine(){
+        String horizontalLine = "____________________________________________________________";
```
EthanYidong: I like the usage of a named variable instead of a magic literal, but could this be a constant instead?

EthanYidong (commented):
Overall the code was reasonably readable with the most common code extracted to functions. However, I noted that the majority of your business logic is still housed under the main function. As your code becomes more complicated, you should consider further extracting out functionality into other classes and methods.

src/main/java/Task.java
```diff
@@ -0,0 +1,22 @@
...
+    protected String description;
+    protected boolean isDone;
+
+    public Task(String description) {
+        this.description = description;
+        this.isDone = false;
+    }
+
+    public String getStatusIcon() {
+        return (isDone ? "X" : " "); // mark done task with X
```
irving11119: Perhaps you could include full documentation (following the Javadoc comments standard) for your class methods beyond just using inline comments.

src/main/java/Task.java
```diff
@@ -0,0 +1,22 @@
...
+    }
+    public void markAsDone(){
```
irving11119: Should this method be named `setIsDone` since it is a setter for the `isDone` class variable?

src/main/java/Alex.java
```diff
@@ -0,0 +1,80 @@
+import java.util.Scanner;
```
irving11119: I like how you chose explicitly which package from `java.util` rather than using a generic import.

src/main/java/Alex.java
```diff
@@ -0,0 +1,80 @@
...
+        System.out.println(greet);
+        //System.out.println(question);
+        printLine();
+    }
+    public static void exit() {
+        String exit = "Bye. Hope to chat with you again soon!";
+        System.out.println(exit);
+        printLine();
+    }
+}
```
irving11119: I think you did a good job at following the naming conventions for variable, class and method names.

irving11119 (commented):
I think your code quality overall is fairly decent and you follow naming conventions well. However, do consider adding in documentation for your methods based on the Javadoc comments standards.

src/main/java/Alex/Alex.java
```diff
@@ -0,0 +1,161 @@
...
+                printLine();
+            } catch (DukeException e) {
+                System.out.println("☹ OOPS!!! I'm sorry, but I don't know what that means :-(");
+                printLine();
+            }
+
+            }
+
+    }
+    public static void echoResponse(Task task, int number) {
```
qianz-z: Remember to add an empty line in between methods!

src/main/java/Alex/Alex.java
```diff
@@ -0,0 +1,161 @@
...
+        String exit = "Bye. Hope to chat with you again soon!";
+        System.out.println(exit);
+        printLine();
+    }
+    /** Method that utilises a taskManager to handle different user inputs accordingly */
+    public static void handleInput(String userInput, TaskManager taskManager) throws DukeException {
+        String[] words = userInput.split(" ");
+        String[] allCommands = new String[]{"todo" , "bye", "event", "list" , "unmark" , "mark", "deadline"};
+        List<String> commandList = new ArrayList<>(Arrays.asList(allCommands));
+        String command = words[0].toLowerCase();
```
qianz-z: Try to avoid magic literals. There has been a few more instances in the rest of your code. You can use a named constant makes the code easier to understand because the name tells us more about the meaning of the number.

src/main/java/Alex/Alex.java
```diff
@@ -0,0 +1,161 @@
...
+            taskManager.getAllTasks()[number - 1].unmark();
+            System.out.println("[" + taskManager.getAllTasks()[number - 1].getStatusIcon() + "]" + " " + taskManager.getAllTasks()[number - 1].getDescription());
+            printLine();
+
+
+        }
+        else {
+            String activity = "";
+            for (int i = 1; i < words.length; i++) {
+                if(words[i].charAt(0) == '/' ) {
```
qianz-z: Try to avoid deep nestings in your code!

src/main/java/Alex/Alex.java
```diff
@@ -0,0 +1,161 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Scanner;
+
+public class Alex {
+    public static void main(String[] args) {
+        TaskManager taskManager = new TaskManager();
+        Scanner myObj = new Scanner(System.in);
+        greeting();
```
qianz-z: Try to keep the abstractions on the same level. Do take a look at how SLAP should be done :>

src/main/java/Alex/Alex.java
```diff
@@ -0,0 +1,161 @@
...
+    }
+    public static void exit() {
+        String exit = "Bye. Hope to chat with you again soon!";
+        System.out.println(exit);
+        printLine();
+    }
+    /** Method that utilises a taskManager to handle different user inputs accordingly */
+    public static void handleInput(String userInput, TaskManager taskManager) throws DukeException {
+        String[] words = userInput.split(" ");
+        String[] allCommands = new String[]{"todo" , "bye", "event", "list" , "unmark" , "mark", "deadline"};
```
qianz-z: You can consider using constants :>

qianz-z (commented):
Overall, good effort in your iP. Do take note of deep nestings and magic literals in your code. Keep it up!

PR #153, opened 2023-01-30
[dendendenden04] iP

dendendenden04:


src/main/java/Psyduck.java
```diff
@@ -0,0 +1,184 @@
...
+        int taskNum;
+        String input = in.nextLine();
+        input = input.trim();
+        int space = input.indexOf(" ");
+        String command;
+        if (space == -1) {
+            command = input;
+        } else {
+            command = input.substring(0, space);
+        }
```
adenteo: Perhaps it is better to modify this into a Boolean function to check whether is it a single/multi-word user input?

src/main/java/Psyduck.java
```diff
@@ -0,0 +1,184 @@
...
+        case "deadline":
+            createDeadline(input, space);
+            break;
+        case "event":
+            createEvent(input, space);
+            break;
+        default:
+            handleInvalidCommand();
+        }
+    }
```
adenteo, line 54: Good use of switch statement for a clean description of each case!

src/main/java/Psyduck.java
```diff
@@ -0,0 +1,184 @@
...
+
+    public static void createDeadline(String input, int space) {
+        if (space == -1) {
+            System.out.println("Deadline task cannot be empty! >:(");
+            linePrint();
+            return;
+        }
+        int byPos = input.indexOf("/by");
+        if (byPos == -1) {
```
adenteo: Perhaps it is better to use Boolean functions for these code segments on input validation? This is applicable to a few other segments as well that I have noted out.

src/main/java/Psyduck.java
```diff
@@ -0,0 +1,184 @@
...
+        if (fromPos == -1 || toPos == -1) {
+            linePrint();
+            System.out.println("Psyduck needs a start or end date.");
+            System.out.println("Remember to use the <deadline> <description> </from> <date> </to> <date> format for your input");
+            linePrint();
+            return;
+        }
+        String description = input.substring(space + 1, fromPos);
+        String from = input.substring(fromPos + 6, toPos);
+        String to = input.substring(toPos + 4);
```
adenteo: Perhaps it is better to avoid magic numbers to make it clearer to the reader? Could consider using named constants or other methods to split the user input.

adenteo (commented):
Overall, I found your code relatively easy to read. Every function was simple and neat and the flow of events were quite clear. I noted a few segments regarding your validation of user inputs which could be quite confusing. There was also a segment where you used magic numbers to obtain substrings of the user input, which may not be intuitive.  Consider using named constants or creating functions for input validation?

src/main/java/Psyduck.java
dendendenden04: Good idea, thank you for your suggestion.

src/main/java/Command.java
```diff
@@ -0,0 +1,210 @@
...
+        System.out.println();
+    }
+
+    public static void createToDo(String input, int space) {
+        if (space == NOT_FOUND) {
+            linePrint();
+            System.out.println("Todo task cannot be empty! >:(");
+            linePrint();
+            return;
+        }
```
okkhoy: Can be handled as an exception

src/main/java/Command.java
```diff
@@ -0,0 +1,210 @@
...
+        int space = input.indexOf(" ");
+        String command;
+        if (space == NOT_FOUND) {
+            command = input;
+        } else {
+            command = input.substring(0, space);
+        }
+        switch (command) {
+        case "bye":
+            //fallthrough
```
okkhoy: 👍

src/main/java/Command.java
```diff
@@ -0,0 +1,210 @@
...
+        linePrint();
+        System.out.println("Invalid command, Psyduck does not understand. (◎_◎;)");
+        System.out.println("Please enter a valid command.");
+        linePrint();
+    }
+
+    public static void processCommands() {
+        String input = in.nextLine();
+        input = input.trim();
+        int space = input.indexOf(" ");
```
okkhoy: Can be extracted to a separate method that does parsing

src/main/java/Psyduck.java
```diff
@@ -0,0 +1,102 @@
...
+
+    public static void listTasks() {
+        Command.linePrint();
+        if (taskCount == 0) { //list is empty
+            System.out.println("List is empty.");
+        } else {
+            for (int i = 0; i < taskCount; i++) {
+                System.out.print(Integer.toString(i + 1) + ".");
+                System.out.println(tasks.get(i));
+            }
```
okkhoy: Make happy path prominent

src/main/java/Storage.java
```diff
@@ -0,0 +1,86 @@
...
+
+public class Storage {
+
+    public static File createFile(String filepath) {
+        try {
+            File f = new File(filepath);
+            if (f.createNewFile()) {
+                System.out.println("File created: " + f.getName());
+            } else {
+            }
```
okkhoy: please do not have empty catch blocks. if you choose, you may just check for the condition in a guard clause

okkhoy (commented):
1. The codebase is not well organized; please complete packages task soon.
2. There is no exception handling yet. 
3. Please include Javadoc comments for all non-trivial methods and classes.

PR #158, opened 2023-01-30
[Hamada Masahiro] iP

Masahiro21:


src/main/java/Duke.java
```diff
@@ -55,7 +55,7 @@ public static Boolean Exit(String args){
         }
     }
      */
-     public static boolean actionCommand(String args){
+     public static boolean shouldExit(String args){
```
chongyongrui: I like how you changed the name of the Boolean variable name to something that sounds natural, and an alternative could have been "isActionCommand"

src/main/java/Duke.java
```diff
@@ -1,20 +1,51 @@
-import java.util.Scanner;
 import java.util.Arrays;
+import java.util.Scanner;
 public class Duke {
-    private static String[] Tasks = new String[100];
+    private static String[] tasks = new String[100];
```
chongyongrui: I feel that you can use a constant variable to replace "100" as it is a magic number

src/main/java/Duke.java
```diff
@@ -25,12 +56,27 @@ public static Boolean Exit(String args){
     }
      */
      public static boolean actionCommand(String args){
-         switch(args){
+         String command = filter(args);
+         int dividerPosition = args.indexOf(" ");
+         String taskNumber = args.substring(dividerPosition + 1, args.length());
+         switch(command){
              case "bye":
```
chongyongrui: I like the use of switch statements instead of if-else statements as your code is neat and clear

src/main/java/Duke.java
```diff
@@ -1,10 +1,100 @@
+import java.util.Arrays;
+import java.util.Scanner;
 public class Duke {
+    private static String[] tasks = new String[100];
+    private static String[] taskStatus = new String[100];
```
cleoncheng2000: Perhaps you can use a variable to store 100? For easy change in the future.

src/main/java/Duke.java
```diff
@@ -1,10 +1,100 @@
...
+
+    public static void listTask(String[] args){
+        for(int i = 1; i <= numOfTask; i+= 1){
+            System.out.println(i + "." + "[" + taskStatus[i-1] + "] "+ args[i-1]);
+        }
+    }
+
+    public static String filter(String sentence){
+        String[] words = sentence.split(" ");
+        for(String word: words){
```
cleoncheng2000: Any reason why you used a loop to just check the first word instead of using the first index?

src/main/java/Duke.java
```diff
@@ -1,10 +1,100 @@
+import java.util.Arrays;
+import java.util.Scanner;
 public class Duke {
+    private static String[] tasks = new String[100];
```
cleoncheng2000: I like how you used "tasks"  for the name of a collection of objects. Good work!

src/main/java/Duke.java
```diff
@@ -1,10 +1,100 @@
...
+    private static String[] taskStatus = new String[100];
+    private static int numOfTask = 0;
+    //Method to Add to the list of task
+    public static void addTask(String args){
```
cleoncheng2000: I like the use of camelCase for the variable name "addTask". Good work!

src/main/java/Duke.java
```diff
@@ -1,10 +1,100 @@
...
+        System.out.println("added: " + args);
+    }
+
+    public static void listTask(String[] args){
+        for(int i = 1; i <= numOfTask; i+= 1){
+            System.out.println(i + "." + "[" + taskStatus[i-1] + "] "+ args[i-1]);
+        }
+    }
+
+    public static String filter(String sentence){
```
cleoncheng2000: Perhaps a more intuitive function name here like "getCommand"?

cleoncheng2000 (commented):
Good work overall!

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,154 @@
...
+import duke.exceptions.*;
+import java.util.Scanner;
+public class Duke {
+    private static Task[] tasks = new Task[100];
+    private static int numOfTask = 0;
+    // Method to list out the tasks stored in the task list
+    public static void listTasks(){
+        System.out.println("Here are the tasks in your list:");
+        for(int i = 1; i <= numOfTask; i+= 1){
+            System.out.println(i + "." +"[" + tasks[i-1].getType() + "]" + "[" + tasks[i-1].getStatusIcon() + "] "+ tasks[i-1].getDescription()
```
slightlyharp: Try to keep line length shorter than 110 characters

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,154 @@
...
+            + tasks[i-1].getDeadline() + tasks[i-1].getPeriod());
```
slightlyharp: Indentation for wrapped lines should be 8 spaces more than the parent line.

src/main/java/duke/Greetings.java
```diff
@@ -0,0 +1,14 @@
+package duke;
+public class Greetings {
```
slightlyharp: Consider dedicate a class for all the output sting instead of creating a class for a single output

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,154 @@
...
+        String toDate = toPeriod.replaceAll("to", "");
+        output[2] = toDate.trim();
+        return output;
+    }
+
+     public static boolean shouldExit(String args) throws DukeException {
+         String command = filterCommand(args);
+         int dividerPosition = args.indexOf(" ");
+         String taskNumber = args.substring(dividerPosition + 1, args.length());  //Used for mark and unmark command
+         switch(command){
```
slightlyharp, line 57: Try to use same level of abstraction for all the cases(eg. either call another method for every case/expand the step for every case)
Since the shouldExit method is quite long calling another method for all the case would be recommended.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,154 @@
...
+        String fromPeriod = deadline.replaceAll("from", "");
+        String fromDate = fromPeriod.substring(0,fromPeriod.indexOf("/"));
+        output[1] = fromDate.trim();
+        String toPeriod = fromPeriod.substring(fromPeriod.indexOf("/")+1, fromPeriod.length());
+        String toDate = toPeriod.replaceAll("to", "");
+        output[2] = toDate.trim();
+        return output;
+    }
+
+     public static boolean shouldExit(String args) throws DukeException {
```
slightlyharp: Consider changing the method name as it is not just determining if the program should exit

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,154 @@
...
+                    return "mark";
+                case "unmark":
+                    return "unmark";
+                case "todo":
+                    return "todo";
+                case "deadline":
+                    return "deadline";
+                case "event":
+                    return "event";
+                default:
```
slightlyharp: There should be no indentation for case clause

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,154 @@
...
+        String toPeriod = fromPeriod.substring(fromPeriod.indexOf("/")+1, fromPeriod.length());
+        String toDate = toPeriod.replaceAll("to", "");
+        output[2] = toDate.trim();
+        return output;
+    }
+
+     public static boolean shouldExit(String args) throws DukeException {
+         String command = filterCommand(args);
+         int dividerPosition = args.indexOf(" ");
+         String taskNumber = args.substring(dividerPosition + 1, args.length());  //Used for mark and unmark command
```
slightlyharp: Since it is only used for mark and unmark shift the statement into the case for better readable/prevent parsing error if the command is not mark/unmark

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,154 @@
...
+        Greetings welcome = new Greetings();
+        System.out.println(welcome);
+        try {
+            Scanner sc = new Scanner(System.in);
+            String in = sc.nextLine();
+            while (shouldExit(in) == false) {
+                in = sc.nextLine();
+            }
+        } catch (ArrayIndexOutOfBoundsException e){
+            System.out.println("This task number does not exist!");
```
slightlyharp: Try to catch all the exception at same level (some exceptions are caught in shouldExit method). Since this catch is only cater to mark command it should be on the same level as those exception caught in shouldExit.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,154 @@
...
+        return sentence;
+    }
+
+    // Method to remove the command word from the input and return the description
+    public static String filterDescription(String sentence) {
+        String command = filterCommand(sentence);
+        String description = sentence.replaceAll(command,"");
+        return description.trim();
+    }
+    // Method to remove the command word, deadline and return the description
```
slightlyharp: Can be more precise on the comments (eg. instead of description, state the task name and deadline date). Same for the other method comments

src/main/java/duke/Deadline.java
```diff
@@ -0,0 +1,22 @@
+package duke;
+public class Deadline extends Task {
+
+    protected String by;
```
slightlyharp, line 8: Try to use noun for the variable. Same for the other classes

src/main/java/duke/Event.java
```diff
@@ -0,0 +1,28 @@
+package duke;
+import duke.exceptions.EventException;
+public class Event extends Task {
+
+    protected String from;
+    protected String to;
```
slightlyharp, line 10: Consider using private as the modifier since it is not called outside the class. Same for the other classes

src/main/java/duke/Event.java
```diff
@@ -0,0 +1,28 @@
...
+
+    public String getType(){
+        return "E";
+    }
+
+    public String getPeriod(){
+        return " (from: " + from + " to: " + to + ")";
+    }
+    @Override
+    public String toString() {
```
slightlyharp, line 43: Consider using a different method name call to return this string since it is only for adding the task and not just the description of the task. Another possible way is to use  to string to return just the description of the task (eg. [E] .... from ... to ...) and at the print add "Got it. I've added this task". to make level of abstraction more consistent

slightlyharp (commented):
Generally well done. Good job for completing the optional task. Just need to take note for the naming of the methods and try to avoid varying the level of abstraction within a code fragment.

PR #159, opened 2023-01-30
[namsengi11] iP

namsengi11:


src/main/java/Duke.java
```diff
@@ -1,10 +1,42 @@
+import java.util.Scanner;
+
 public class Duke {
+    private final static String LINE_BREAK = " ____________________________________________________________";
```
YC-Michael: Good usage of constants with the uppercase naming

src/main/java/Duke.java
```diff
@@ -1,10 +1,42 @@
...
-                + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
+
+        Scanner scanner = new Scanner(System.in);
+        speak(GREETING);
+        boolean running = true;
+        while (running) {
+            String command = scanner.nextLine();
+            String[] commandByWord = command.split(" ");
+            if (command.equals("bye")) { // whats the difference between equals and ==
```
YC-Michael: .equals() if two objects have the same content while == checks if two objects (non-primitive data type) point to the same address

src/main/java/Duke.java
```diff
@@ -1,10 +1,42 @@
...
+                speak(EXITING);
+                running = false;
+            } else if (command.equals("list")) {
```
YC-Michael: Good nesting

src/main/java/Duke.java
```diff
@@ -1,10 +1,42 @@
...
+                speak(tasks.listTasks());
+            } else if (commandByWord[0].equals("mark")) {
+                speak(tasks.mark(Integer.parseInt(commandByWord[1])));
```
YC-Michael: Avoid the use of complicated expression and try to split up the code into intermediate steps.
```suggestion
                Integer taskIndex = Integer.parseInt(commandByWord[1]);
                String messageToUser = tasks.mark(taskIndex);
                speak(messageToUser);
```

YC-Michael (commented):
Overall, good encapsulation in Tasks class by limiting access to the taskList. Good naming and readability.

src/main/java/Duke/Duke.java
```diff
@@ -0,0 +1,49 @@
+package Duke;
+
+import java.util.Scanner;
+
+public class Duke {
+    private final static String LINE_BREAK = " ____________________________________________________________";
```
jeyvia, line 11: Good job using constants for your strings! You could extend that to more magic strings (e.g `"bye"`) as well as magic numbers!

src/main/java/Duke/Duke.java
```diff
@@ -0,0 +1,49 @@
...
+                speak(tasks.addTask(commandByWord));
+            } else if (commandByWord[0].equals("delete")) {
+                speak(tasks.deleteTask(commandByWord));
+            } else {
+                speak(ERR_MESSAGE);
+            }
+        }
+    }
+
+    static void speak(String message) {
```
jeyvia: Do give your method here either public or private based on what you think is suitable :)

jeyvia: This applies to your other methods that lack either public or private as well.

src/main/java/Duke/Event.java
```diff
@@ -0,0 +1,66 @@
...
+            if (index != indexOfFrom) {
+                startDate += " ";
+            }
+
+            startDate += commandByWord[index];
+            ++index;
+        }
+
+        for (int i = index + 1; i < commandByWord.length; ++i) {
+            // add space between words
```
jeyvia: Please remove such comments as they go against the java coding standards.

src/main/java/Duke/TaskCreator.java
```diff
@@ -0,0 +1,24 @@
...
+
+public class TaskCreator {
+    public static Task createNewTask(String[] commandByWord)
+            throws IllegalArgumentException, ArrayIndexOutOfBoundsException {
+        String type = commandByWord[0];
+        if (type.equals("todo")) {
+            return ToDo.createToDo(commandByWord);
+        } else if (type.equals("deadline")) {
+            return Deadline.createDeadline(commandByWord);
+        } else {
```
jeyvia: You should not default your command behaviours like this! Create a `if(condition)` for each command :)

src/main/java/Duke/TaskUpdater.java
```diff
@@ -0,0 +1,117 @@
...
+            File file = new File("data/duke.txt");
+            List<String> lines = Files.lines(file.toPath())
+                    .map(line ->
+                            line.substring(0, line.length() - charAtEnd)
+                                    .equals(taskWithoutCompletionStatus) ?
+                                    task :
+                                    line
+                            )
+                    .collect(Collectors.toList());
+//            for (String line : lines) {
```
jeyvia: Instead of leaving it commented out, you can remove it :)

src/main/java/Duke/Duke.java
jeyvia: You could also extend it to your booleans where possible!

PR #172, opened 2023-01-30
[Yat Pang] iP

YatPang:


src/main/java/Duke.java
```diff
@@ -1,10 +1,129 @@
+import java.util.Arrays;
+import java.util.Scanner;
 public class Duke {
+
+    public static Task[] addTask(Task[] tasks, Task task){
```
LimHongYao: maybe you can have an alternate naming for `task` since 6/8 words are some variation of "task"

src/main/java/Duke.java
```diff
@@ -1,10 +1,129 @@
...
+        for (int i = idx; i < strings.length; i++){
+            if (strings[i] == null){
+                break;
+            } else {
+                joinedString += strings[i] + " ";
+            }
+        }
+        return joinedString.trim();
+    }
+    public static void welcome(){
```
LimHongYao: maybe `welcomeMessage` is a better name? Since it does not seem to have any other function

src/main/java/Duke.java
```diff
@@ -1,10 +1,129 @@
...
+        } else {
+            tasks[Integer.parseInt(strings[1]) - 1].markAsNotDone();
+            System.out.println("OK, I've marked this task as not done yet:");
+            Task tsk = tasks[Integer.parseInt(strings[1]) - 1];
+            System.out.println(printIconStatus(tsk));
+        }
+    }
+
+    public static void mark(String[] strings, Task[] tasks){
+        if (Integer.parseInt(strings[1]) > tasks.length){
```
LimHongYao: good handling of out of bounds!

src/main/java/Events.java
```diff
@@ -0,0 +1,16 @@
+public class Events extends Task{
+    protected String start;
+    protected String end;
+    public Events(String description, String start, String end) {
+        super(description);
+        this.start = start;
+        this.end = end;
+    }
+    public String getIcon(){
```
LimHongYao, line 18: is this an override from` Task` ? Potentially buggy otherwise as you might call the wrong `getIcon()` i think

LimHongYao (commented):
gj bro quite neat but you might want to use more explicit naming

src/main/java/Duke.java
jeyvia: You could rename your array of Task(s) to `TaskList` or something like that to avoid confusion. Currently you are overusing the word "task"

src/main/java/Duke.java
```diff
@@ -1,10 +1,129 @@
...
+        } else {
+            tasks[Integer.parseInt(strings[1]) - 1].markAsDone();
+            System.out.println("Nice! I've marked this task as done:");
+            Task tsk = tasks[Integer.parseInt(strings[1]) - 1];
+            System.out.println(printIconStatus(tsk));
+        }
+    }
+
+    public static String printIconStatus(Task tsk){
+        // if (tsk instanceof [Class]) to check if tsk is of a certain class
```
jeyvia: Please remove such comments as they go against the java coding standards.

src/main/java/ToDos.java
```diff
@@ -0,0 +1,8 @@
+public class ToDos extends Task{
+    public ToDos(String description) {
+        super(description);
+    }
+    public String getIcon(){
```
jeyvia: Remember to leave a space between `()` and `{`. There are multiple instances of that throughout your code. Please remember to fix it  :)

src/main/java/Task.java
```diff
@@ -0,0 +1,28 @@
...
+        isDone = true;
+    }
+
+    public void markAsNotDone(){isDone = false;}
+
+    public String getIcon(){
+        return " ";
+    }
+
+    public String getStart(){return "";}
```
jeyvia: Even though these are simple 1 liners, please do still follow the proper style (aka Egyptian Style).

src/main/java/Duke.java
```diff
@@ -1,10 +1,129 @@
...
-                + "|  _ \\ _   _| | _____ \n"
-                + "| | | | | | | |/ / _ \\\n"
-                + "| |_| | |_| |   <  __/\n"
-                + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
+        welcome();
+        String input;
+        Task[] tasks = new Task[0];
+        int lstID = 1;
+        while (true) {
```
jeyvia, line 28: I notice that you have a tendency to use magic strings and numbers. Do try to avoid that where possible, and instead use constants to set your strings/numbers.

PR #177, opened 2023-01-31
[Tay Jiun Yuan] iP

TayJiunYuan:


src/main/java/Task.java
```diff
@@ -0,0 +1,26 @@
...
+    public void markAsDone(){
+        this.isDone = true;
+    }
+
+    public void markAsUndone(){
+        this.isDone = false;
+    }
+    
+    public String printTask(){
+        return description;
```
Thiolk: Perhaps you could consider keeping the space between the "()" and "{" consistent?

src/main/java/Duke.java
```diff
@@ -1,10 +1,124 @@
...
+                break;
+            case "todo":
+                addTodo(userInput);
+                break;
+            case "deadline":
+                addDeadline(userInput);
+                break;
+            case "event":
+                addEvent(userInput);
+                break;
```
Thiolk: Maybe you could consider adding a default branch in the case statements?

src/main/java/Duke.java
```diff
@@ -1,10 +1,124 @@
...
+    }
+
+    public static void displayTaskList(){
+        printLine();
+        for (int i = 0; i < tasks.size(); i++){
+            String taskNumber = Integer.toString(i+1);
+            Task currentTask = tasks.get(i);      
+            System.out.println("\t" + taskNumber + "." +currentTask.getStatusIcon() + currentTask.printTask());
+        
+        }
```
Thiolk: Any reason why you included the extra empty line at the end of the for loop?

src/main/java/Duke.java
```diff
@@ -1,10 +1,124 @@
...
+        printLine();
+        System.out.println("\t"+'"'+currentTask.description+'"'+" is marked as done! Good Job :)" );
+        printLine();
+    }
+
+    public static void unmarkTask(int index){
+        Task currentTask = tasks.get(index-1); 
+        currentTask.markAsUndone();
+        printLine();
+        System.out.println("\t"+'"'+currentTask.description+'"'+" is marked as undone. Jiayou!" );
```
Thiolk: Perhaps you could include some spaces between the " in the println statments to improve the readability?

src/main/java/Duke.java
```diff
@@ -1,10 +1,124 @@
...
+        System.out.println( "\tThank you! See you again soon :)");
+        printLine();
+    }
+
+    public static void addEvent(String userInput){
+        int fromIndexStart = userInput.indexOf(" /from");
+        int fromIndexEnd = fromIndexStart + 7;
+        int toIndexStart = userInput.indexOf(" /to");
+        int toIndexEnd = toIndexStart + 5;
+        int startIndex = 5;
```
Thiolk: Should you avoid the use of magic numbers and replace the number with a named constant instead?

Thiolk (commented):
Overall, the code was done to a high standard and was easily readable though there are certain sections that could be improved slightly to keep the style of your code more consistent

src/main/java/Duke.java
MichelleLiang0116: Try to avoid using constant, probably can use replace or substring command.

src/main/java/Duke.java
```diff
@@ -24,11 +24,40 @@ public static void printGoodbye(){
...
+        System.out.println("\tadded: " + eventDescription);
+        printLine(); 
+    } 
 
-    public static void addToList(String userInput){
-        tasks.add(new Task(userInput));
+    public static void addDeadline(String userInput){
+        int byIndexStart = userInput.indexOf(" /by");
+        int byIndexEnd = byIndexStart + 4;
+        int startIndex = 9;
```
MichelleLiang0116: Try to avoid using constant, probably can use replace or substring command.

src/main/java/Duke.java
```diff
@@ -24,11 +24,40 @@ public static void printGoodbye(){
...
+        tasks.add(new Deadline(deadlineDescription, deadlineDate));
         printLine();
-        System.out.println("\tadded: " + userInput);
+        System.out.println("\tadded: " + deadlineDescription);
+        printLine(); 
+    }
+
+
+    public static void addTodo(String userInput){
+        int startIndex = 4;
```
MichelleLiang0116: Try to avoid using constant, probably can use index of or substring command.

src/main/java/Duke.java
```diff
@@ -80,8 +109,14 @@ public static void main(String[] args) {
...
                 break;
-            default:
-                addToList(userInput);
+            case "todo":
+                addTodo(userInput);
+                break;
+            case "deadline":
+                addDeadline(userInput);
+                break;
+            case "event":
```
MichelleLiang0116: Probably consider case sensitive problem.

MichelleLiang0116 (commented):
Overall is good.
Just try not to use constant in order to be more flexible.

src/main/java/Exceptions/InvalidCommandException.java
```diff
@@ -0,0 +1,5 @@
+package exceptions;
+
+public class InvalidCommandException extends Exception{
+
```
R-Ramana: Don't have empty classes, same for the other files

src/main/java/commands/Command.java
```diff
@@ -0,0 +1,130 @@
...
+        throw new InvalidCommandException();
+    }
+
+    public static boolean isEventValid(String[] userInputArray){
+        if (userInputArray.length == 1){
+            return false;           //return false if task is empty
+        }
+        boolean fromExists = false;
+        boolean toExists = false;
+        boolean descriptionAndDatesExists = false;
```
R-Ramana: something like `containsFrom` or `hasFrom` etc would work betetr as it makes it more readable

PR #178, opened 2023-01-31
[nafisazizir] iP

nafisazizir:


src/main/java/Deadline.java
```diff
@@ -0,0 +1,17 @@
+public class Deadline extends Task {
+    private final String by;
```
Reynold-SL:
```suggestion
    private final String BY
```

perhaps you could try changing the naming of the variable to the above? I notice similar issues in other places

src/main/java/Jonathan.java
```diff
@@ -30,24 +30,59 @@ public static void main(String[] args) {
...
+                String description = command.substring(command.indexOf(" ") + 1, command.indexOf(" /from"));
+                String start = command.substring(command.indexOf("/from") + 6, command.indexOf(" /to"));
+                String end = command.substring(command.indexOf("/to") + 4);
+                Event event = new Event(description, start, end);
+                list[taskCounter] = event;
                 taskCounter += 1;
-                System.out.println(line + "\n    added: " + command + "\n" + line + "\n");
+                addedMessage(event, taskCounter);
+
+            } else { // add Task
```
Reynold-SL:
```suggestion
   } else { 
       // add task
```

Perhaps would it be better if you indent the message above the position in the code? I notice a similar issue in other places

src/main/java/Jonathan.java
```diff
@@ -30,24 +30,59 @@ public static void main(String[] args) {
...
+                System.out.println("    ____________________________________________________________\n" +
+                        "    Wrong input, please type it correctly!\n" +
+                        "    ____________________________________________________________\n");
             }
         }
     }
 
+    /*
+    Method to print the added message to the user after prompting input
```
Reynold-SL:
```suggestion
   * Method to print the added message to the user after prompting input
```

Perhaps you could add a * in front of this comment? I notice a similar issue in other places.

src/main/java/Jonathan.java
```diff
@@ -30,24 +30,59 @@ public static void main(String[] args) {
...
+
+            } else if (command.split(" ")[0].equals("event")) {
+                String description = command.substring(command.indexOf(" ") + 1, command.indexOf(" /from"));
+                String start = command.substring(command.indexOf("/from") + 6, command.indexOf(" /to"));
+                String end = command.substring(command.indexOf("/to") + 4);
+                Event event = new Event(description, start, end);
+                list[taskCounter] = event;
                 taskCounter += 1;
-                System.out.println(line + "\n    added: " + command + "\n" + line + "\n");
+                addedMessage(event, taskCounter);
```
Reynold-SL: I like the naming convention you have used here

src/main/java/Jonathan.java
```diff
@@ -30,24 +30,59 @@ public static void main(String[] args) {
                 boolean isDone = command.split(" ")[0].equals("mark");
                 int taskNum = Integer.parseInt(command.split(" ")[1]);
                 Task task = list[taskNum-1];
-
                 changeStatus(task, isDone);
 
-            } else { // add Task
```
Reynold-SL: I like the indentations you have used here

src/main/java/Task.java
```diff
@@ -0,0 +1,29 @@
...
+
+    public String getDescription() {
+        return description;
+    }
+
+    public void setDescription(String description) {
+        this.description = description;
+    }
+
+    public void setDone(boolean done) {
```
Reynold-SL, line 18:
```suggestion 
 public void setDone(boolean isDone) {
```

Perhaps you can change the naming of the boolean input to the above?

Reynold-SL:
```suggestion 
 public void setDone(boolean isDone) {
```

Perhaps you can change the naming of the boolean input to the above?

src/main/java/Jonathan.java
```diff
@@ -0,0 +1,110 @@
...
+public class Jonathan {
+    public static void main(String[] args) {
+        Scanner input = new Scanner((System.in));
+        String command;
+        String line = "    ____________________________________________________________";
+        String welcome = line + "\n    Hello! I'm Jonathan\n" +
+                "    What can I do for you?\n" + line + "\n";
+        String goodbye = "    Bye. Hope to see you again soon!\n" + line;
+
+        Task[] list = new Task[100];
```
BernardLesley: It may be clearer to declare a constant (for example: TASK_LENGTH) to denote the length of array

src/main/java/Jonathan.java
```diff
@@ -0,0 +1,110 @@
...
+                taskCounter += 1;
+                addedMessage(event, taskCounter);
+
+            } else { // add Task
+                System.out.println("    ____________________________________________________________\n" +
+                        "    Wrong input, please type it correctly!\n" +
+                        "    ____________________________________________________________\n");
+            }
+        }
+    }
```
BernardLesley: I think you can consider using switch case, instead of if else. However the readability is still good in this case.

src/main/java/Jonathan.java
BernardLesley: You can consider changing the naming of 'list' to 'tasks', because the naming may be too vague

src/main/java/Jonathan.java
```diff
@@ -0,0 +1,110 @@
...
+                    "    ____________________________________________________________\n");
+        } else { // Unmark the Task
+            System.out.println("    ____________________________________________________________\n" +
+                    "     OK, I've marked this task as not done yet:\n" +
+                    "       " + task + "\n" +
+                    "    ____________________________________________________________\n");
+        }
+    }
+
+}
```
BernardLesley: I like how you extract and group the method

BernardLesley (commented):
Overall, the code has adhered to the coding standards and code quality guidelines. Some minor fixes and you're good to go. Well done!

src/main/Java/jonathan/Jonathan.java
```diff
@@ -0,0 +1,257 @@
...
+                } else if (task instanceof Deadline) {
+                    Deadline deadline = (Deadline) task;
+                    String line = "D | " + deadline.getIsDone() + " | " + deadline.getDescription() +
+                            " | " + deadline.getBy() + "\n";
+                    outFileWriter.write(line);
+                } else if (task instanceof Event) {
+                    Event event = (Event) task;
+                    String line = "E | " + event.getIsDone() + " | " + event.getDescription() +
+                            " | " + event.getStart() + " | " + event.getEnd() + "\n";
+                    outFileWriter.write(line);
```
aditichadha1310: Try to avoid the usage of magic strings as much as possible. For example: in this part of code, you are using the pipe symbol multiple times. You can consider saving it in a variable and use that instead.

src/main/Java/jonathan/Jonathan.java
```diff
@@ -0,0 +1,257 @@
...
+            File file = new File("data/jonathan.txt");
+
+            Scanner reader = new Scanner(file);
+
+            while (reader.hasNextLine()) {
+                String line = reader.nextLine();
+                String[] extract = line.split("\\|");
+                String symbol = extract[0].trim();
+                String isDone = extract[1].trim();
+                String description = extract[2].trim();
```
aditichadha1310: Try to avoid the usage of Magic numbers here.

src/main/Java/jonathan/Jonathan.java
```diff
@@ -0,0 +1,257 @@
...
+    /*
+    Method to print all the Task, including the Task's status
+     */
+    public static void lookupTask(ArrayList<Task> list, int counter) {
+        System.out.println("    ____________________________________________________________\n" +
+                "    Here are the tasks in your list:");
+        for (int i = 0; i < counter; i++) {
+            System.out.println("    " + (i+1) + ". "  + list.get(i));
+        }
+        System.out.println("    ____________________________________________________________\n");
```
aditichadha1310: Since, you're printing the string "__________________________________" multiple times, you should store it in a variable or create a separate function to print it and use that instead. It is encouraged that no piece of code repeats itself in the files. Refer to the "DRY" Principle.

src/main/Java/jonathan/Jonathan.java
```diff
@@ -0,0 +1,257 @@
...
+            try {
+                if (Objects.equals(command, "bye")) {
+                    saveData(list);
+                    System.out.println(goodbye);
+                    break; // quit the program
+
+                } else if (Objects.equals(command, "list")) {
+                    lookupTask(list, taskCounter);
+
+                } else if (command.split(" ")[0].equals("mark") || command.split(" ")[0].equals("unmark")) {
```
aditichadha1310: This is a fairly complex statement if code readability is taken to account. To improve this readability, you should consider simplifying it by referring to the coding readability standards. Link - https://nus-cs2113-ay2122s1.github.io/website/se-book-adapted/chapters/codeQuality.html#avoid-complicated-expressions

aditichadha1310 (commented):
Very good job overall!

PR #183, opened 2023-01-31
[Ansenn] iP

Ansenn:


src/main/java/Duke.java
```diff
@@ -6,5 +9,36 @@ public static void main(String[] args) {
                 + "| |_| | |_| |   <  __/\n"
                 + "|____/ \\__,_|_|\\_\\___|\n";
         System.out.println("Hello from\n" + logo);
+
+        System.out.println("Hello! I'm Duke\n");
+        System.out.println(" What can I do for you?\n");
+
+        String array[] = new String[100];
```
avielcx: It will be good to eliminate magic numbers and replace them with constants

src/main/java/Duke.java
```diff
@@ -6,5 +9,36 @@ public static void main(String[] args) {
...
+        int counter = 0; // number of items in the list
```
avielcx: It will be good to replace your variable number to be more specific

src/main/java/Duke.java
avielcx: It will be good to rename your array to what the array represents

avielcx:
```suggestion
        String[] array = new String[100];
```
Your array declaration is C-style and not Java

src/main/java/Duke.java
```diff
@@ -6,5 +9,36 @@ public static void main(String[] args) {
...
+
+            if (input.equals("bye")) {
+                isExit = true;
+                System.out.println("______________\n");
+                System.out.println("Bye. Hope to see you again soon!\n");
+                System.out.println("______________\n");
+            } else if (input.equals("list")) {
+                System.out.println("______________\n");
+                for (int i = 0; i < counter; i++) {
+                    System.out.println((i + 1) + ". " + array[i]);
```
avielcx: It will be good to eliminate magic number

src/main/java/Duke.java
```diff
@@ -6,5 +9,36 @@ public static void main(String[] args) {
                 + "| |_| | |_| |   <  __/\n"
                 + "|____/ \\__,_|_|\\_\\___|\n";
         System.out.println("Hello from\n" + logo);
+
+        System.out.println("Hello! I'm Duke\n");
+        System.out.println(" What can I do for you?\n");
+
+        String array[] = new String[100];
+        int counter = 0; // number of items in the list
+        boolean isExit = false;
```
avielcx: It will be good to replace the variable name for greater readability. It takes time for readers to understand what is the variable for (exit from what?).

src/main/java/Duke.java
```diff
@@ -1,10 +1,76 @@
...
+                + "| /    | |    /  _  \\ | |\\  /| |\n"
+                + "| \\ __ | |___ | |_| | | | \\/ | |\n"
+                + "\\____/ |____/ \\_____/ |_|    |_|\n";
+
         System.out.println("Hello from\n" + logo);
+
+        System.out.println("Hello! I'm CLoM!\n");
+        System.out.println(" What can I do for you?\n");
+
+        int counter = 0; // number of items in the list
```
R-Ramana: Name is too generic, something more specific like `xCounter` where x is line or input or task depending on your use case would work better

PR #184, opened 2023-01-31
[tangphi] iP

tangphi:


src/main/java/Task.java
```diff
@@ -0,0 +1,15 @@
+public class Task {
+    protected String description;
+    protected boolean isDone;
```
jingyaaa: boolean variable named properly, good job

src/main/java/Duke.java
```diff
@@ -1,10 +1,87 @@
...
+
 public class Duke {
+
+    //TODO make methods for the different operations
+    protected static Task[] list = new Task[100];
+    public static final String OPENING_MSG = "Hello! I'm Dukebot\n\tWhat can I do for you?";
+    public static final String CLOSING_MSG = "Goodbye! Hope to see you again soon ^^!";
+    public static final String HORIZONTAL = "---------------------------------";
+    public static final String MARK_MSG = "Nice! I've marked this task as done:";
+    public static final String UNMARK_MSG = "Oki! I've marked this task as not done yet:";
```
jingyaaa: constant names are in the correct format :)

src/main/java/Duke.java
```diff
@@ -1,10 +1,87 @@
...
+                StatusToggle(line, item, MARK_MSG, true);
+
+            }
+
+            // Unmark item: constraint - user input begins with "mark"
+            if (line.startsWith("unmark")) {
+
+                StatusToggle(line, item, UNMARK_MSG, false);
+
+            }
```
jingyaaa: can perhaps consider "else if" statements?

src/main/java/Task.java
```diff
@@ -0,0 +1,15 @@
...
+    public Task(String description) {
+        this.description = description;
+        this.isDone = false;
+    }
+    public String getStatusIcon() {
+
+        return (isDone ? "X" : " "); // mark done task with X
+    }
+    
+}
```
jingyaaa: all variables and functions are properly named :)

src/main/java/Duke.java
```diff
@@ -1,10 +1,87 @@
+import java.util.Scanner;
+import java.util.Arrays;
```
jingyaaa: Imported classes are listed explicitly, good job

jingyaaa (commented):
Generally no issues with coding standard, well done

src/main/java/Duke.java
```diff
@@ -1,10 +1,87 @@
...
+
+        Scanner in = new Scanner(System.in);
+        String line = in.nextLine().toLowerCase();
+
+        int listNum;
+
+        int item = 0;
+
+        while (!line.equals("bye")) {
+            if (!(line.equals("list") || line.startsWith("mark") || line.startsWith("unmark"))) {
```
nichyjt: Perhaps this if statement can be refactored out to the bottom as a "default" case. Further, the condition expression is quite long! 

In the code below, checks have already been accounted for:
`list, mark, unmark`.

Since all 3 commands have been checked already, the condition evaluated in this line:  
`if (!(line.equals("list") || line.startsWith("mark") || line.startsWith("unmark"))`  

Is implicitly satisfied from line 58 onwards. So, you can consider factoring out the logic for the addition of a new Task to line 58!

src/main/java/Duke.java
```diff
@@ -1,10 +1,87 @@
...
+
+    }
+
+    public static void StatusToggle(String line, int item, String msg, boolean status) {
+        String lineAr[] = new String[2];
+
+        lineAr = line.split(" ");
+        int listNum = Integer.parseInt(lineAr[1]) - 1;
+
+        if (listNum >= 0 && listNum < item) {
```
nichyjt: The logic here makes sense and accounts for invalid input, good job! 

Perhaps it is possible to build on this logic to reduce the nesting and [make the execution path even clearer](https://nus-cs2113-ay2223s2.github.io/website/se-book-adapted/chapters/codeQuality.html#make-the-happy-path-prominent). 

I think that you can consider refactoring the error check for `listNum` as a guard clause and then doing the `isDone` check.

Possible pseudocode:
```java
// Check for listNum validity
if(listNum < 0 || listNum >= item) {
    // Handle case where item does not exist
}
// Check for done status
if(list[listNum.isDone != status) {
    // Do something
}else{
    // Do something else
}


```

nichyjt (commented):
The code flows logically and is easy to read. There are proper checks for invalid input implemented, which is well thought of!

To build upon this, there are some refactors commented that can be done to further improve readability and the structure of the code that you can consider adopting!

src/main/java/Task.java
tangphi: Thanks for the feedback :)

src/main/java/Duke.java
tangphi: Thanks will consider refactoring that in my code!

src/main/java/Duke.java
```diff
@@ -1,10 +1,87 @@
...
+
+                System.out.println(HORIZONTAL + "\n\t" + msg);
+                System.out.println("\n\t\t" + "[" + list[listNum].getStatusIcon() + "] " + list[listNum].description + "\n");
+                System.out.println(HORIZONTAL);
+            } else {
+                System.out.println("No change, task was already as is");
+            }
+
+        } else {
+            System.out.println("Item number " + (listNum + 1) + " does not exist yet");
```
siyuancheng178: Please separate into different methods to handle different cases, such as `displayAllTasks()`, `UnmarkTask()` etc.

src/main/java/Duke.java
```diff
@@ -1,10 +1,87 @@
...
+            }
+
+        }
+
+        System.out.println(HORIZONTAL + "\n\t" + CLOSING_MSG + "\n" + HORIZONTAL);
+
+    }
+
+    public static void StatusToggle(String line, int item, String msg, boolean status) {
+        String lineAr[] = new String[2];
```
siyuancheng178: Maybe consider renaming this variable? It looks a bit confusing to me

src/main/java/Duke.java
```diff
@@ -1,10 +1,87 @@
...
+            if (line.startsWith("mark")) {
+
+                StatusToggle(line, item, MARK_MSG, true);
+
+            }
+
+            // Unmark item: constraint - user input begins with "mark"
+            if (line.startsWith("unmark")) {
+
+                StatusToggle(line, item, UNMARK_MSG, false);
```
siyuancheng178: Please rename it to `toggleStatus()`. Method should look like a verb and start with lowercase

PR #188, opened 2023-01-31
[Guillaume Gerony] iP

guillaume-grn:


src/main/java/Duke.java
```diff
@@ -48,61 +48,71 @@ public static void main(String[] args) {
...
-                System.out.println("\t" + commands_list[next_index - 1]);
+                printSpecificTask(nextIndexInList - 1, commands, "Got it. I've added this task:");
                 printDashLine();
             } else if (line.length() > 5 && line.substring(0, 5).equals("event")) {
-                int index_from = line.indexOf("/");
-                int index_to = line.indexOf("/", index_from + 1);
-                commands_list[next_index] = new Event(line.substring(0, index_from - 1),
-                        line.substring(index_from + 6, index_to - 1), line.substring(index_to + 4));
-                next_index++;
+                int indexFrom = line.indexOf("/");
```
ollayf: perhaps there could be other ways to do this that are more readable and less susceptible to mistakes. For example, using the <string instance>.split function.

src/main/java/Duke.java
```diff
@@ -48,61 +48,71 @@ public static void main(String[] args) {
...
+                int indexTo = line.indexOf("/", indexFrom + 1);
+                commands[nextIndexInList] = new Event(line.substring(0, indexFrom - 1),
+                        line.substring(indexFrom + 6, indexTo - 1), line.substring(indexTo + 4));
+                nextIndexInList++;
                 printDashLine();
-                System.out.println("\t Got it. I've added this task:");
-                System.out.println("\t" + commands_list[next_index - 1]);
+                printSpecificTask(nextIndexInList - 1, commands, "Got it. I've added this task:");
                 printDashLine();
             } else if (line.length() > 8 && line.substring(0, 8).equals("deadline")) {
```
ollayf: It's still ok... but again would be better if you used string.split function

src/main/java/Duke.java
```diff
@@ -5,23 +5,23 @@
 import util.Deadline;
```
ollayf: Good use of imports and splitting into different files!

ollayf (commented):
Not bad! You can even use the different stuff learnt from coding exercises

src/main/java/Duke.java
```diff
@@ -1,10 +1,126 @@
...
+                int indexTo = line.indexOf("/", indexFrom + 1);
+                commands[nextIndexInList] = new Event(line.substring(6, indexFrom - 1),
+                        line.substring(indexFrom + 6, indexTo - 1), line.substring(indexTo + 4));
+                nextIndexInList++;
+                printDashLine();
+                printSpecificTask(nextIndexInList - 1, commands, "Got it. I've added this task:");
+                printDashLine();
+            } else if (line.split(" ")[0].equals("deadline")) {
+                int index_by = line.indexOf("/");
+                commands[nextIndexInList] = new Deadline(line.substring(9, index_by - 1), line.substring(index_by + 4));
```
siyuancheng178: Maybe you can try another way of parsing strings to avoid these magic numbers here?

src/main/java/Duke.java
```diff
@@ -1,10 +1,126 @@
...
+            if (line.equals("bye")) {
+                printByeMessage();
+                not_finished = false;
+            } else if (line.equals("list")) {
+                displayCommandsList(commands);
+            } else if (line.split(" ")[0].equals("mark")) {
+                int index = Integer.parseInt(line.split(" ")[1]);
+                commands[index].setDone(true);
+                printDashLine();
+                printSpecificTask(index, commands, "Nice! I've marked this task as done:");
```
siyuancheng178: Perhaps you can use a constant variable to avoid the magic string here

PR #201, opened 2023-01-31
[LimHongYao] iP

LimHongYao:
Renamed Duke
Added ToDoList functionality (add, mark, unmark, show list)
Given some personality

### DIFF
--- a/src/main/java/Duke.java
+++ b/src/main/java/Duke.java
@@ -1,8 +1,4 @@
-import java.io.FileWriter;
-import java.io.IOException;
-import java.util.ArrayList;
 import java.util.Scanner;
-import java.io.File;
 
 public class Duke {
     static final int COMMAND_INDEX = 0;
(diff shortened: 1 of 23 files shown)


src/main/java/Duke.java
```diff
@@ -1,10 +1,75 @@
...
-                + "| |_| | |_| |   <  __/\n"
-                + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
+        System.out.println("Hi it's Anna!\nWhat do you need to do?");
+        Scanner in = new Scanner(System.in);
+
+        boolean exit = false;
+        while (!exit) {
+            String input = (in.nextLine()).trim();
+            String inputCMD, inputItem;
```
khooyourun: Can I check what does CMD mean? If it stands for command. I think it's better to just name it as inputCommand for readibilty.

src/main/java/Duke.java
```diff
@@ -1,10 +1,75 @@
...
+            if (input.contains(" ")) {
+                inputCMD = input.split(" ", 2)[0];
+                inputItem = input.split(" ", 2)[1];
```
khooyourun: The numbers like 2, 0, 1 might be considered magic numbers that hinder readability.

src/main/java/ToDoList.java
```diff
@@ -0,0 +1,42 @@
...
+    }
+    public static void viewList () {
+        for (int i = 0; i < TaskList.size(); ++i) {
+            System.out.print(i+1 + ". ");
+
+            System.out.println(TaskList.get(i).getTask());
+        }
+    }
+    public static void markDone (int index) {
+        if (TaskList.get(index).getStatusIcon().equals(" ")) {
```
khooyourun: Is this if statement necessary? I think markDone would change the status icon to "X" independent of what is originally is

src/main/java/Duke.java
```diff
@@ -1,10 +1,75 @@
...
+            case "mark":
+                if (inputItem == null) {
+                    System.out.println("What should I mark?");
+                    ToDoList.viewList();
+                    inputItem = in.nextLine().trim();
+                }
+
+                ToDoList.markDone(Integer.parseInt(inputItem)-1);
+                System.out.println("Okay I've marked item " + inputItem + " as done:");
+                ToDoList.printItem(Integer.parseInt(inputItem)-1);
```
khooyourun: I like the single level of abstraction applied here. Makes things easy to follow

src/main/java/Duke.java
```diff
@@ -1,10 +1,75 @@
...
-        String logo = " ____        _        \n"
-                + "|  _ \\ _   _| | _____ \n"
-                + "| | | | | | | |/ / _ \\\n"
-                + "| |_| | |_| |   <  __/\n"
-                + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
+        System.out.println("Hi it's Anna!\nWhat do you need to do?");
+        Scanner in = new Scanner(System.in);
+
+        boolean exit = false;
```
khooyourun: Preferably name boolean in such a way that implies it is boolean.

khooyourun (commented):
Overall, the code is readable and the if statements are not deeply nested.

src/main/java/Duke.java
liuziyang020319: Maybe use isExit to replace exit can imply a boolean

src/main/java/ToDoList.java
```diff
@@ -0,0 +1,42 @@
+import java.util.ArrayList;
+
+public class ToDoList {
+    private static final ArrayList<Task> TaskList = new ArrayList<>(10);
```
liuziyang020319: Maybe V\variable names need to be in camelCase.

liuziyang020319: Maybe variable names need to be in camelCase.

src/main/java/ToDoList.java
```diff
@@ -0,0 +1,42 @@
...
+    private static int NumTasks = 0;
```
liuziyang020319: Maybe variable names must be in camelCase, same problem.

src/main/java/ToDoList.java
liuziyang020319: Maybe the size of the ArrayList should be a magic number.

liuziyang020319 (commented):
Most the codes follow the coding standard

src/main/java/Duke.java
```diff
@@ -1,10 +1,143 @@
+import java.lang.reflect.Array;
+import java.util.ArrayList;
+import java.util.Scanner;
+
 public class Duke {
+    static final int COMMAND_INDEX = 0;
```
jeyvia: Good job using constants for your numbers! You could extend that to magic strings as well!

src/main/java/Duke.java
```diff
@@ -1,10 +1,143 @@
...
+                return;
+            case "list":
+                if (TaskList.getNumItems() == 0) {
+                    System.out.println("We are free! Let's go play!");
+                } else {
+                    System.out.println("Here's what we've gotta do:");
+                    TaskList.viewList();
+                }
+                break;
+
```
jeyvia: Do avoid leaving the space between each `case` as per the Java Coding Standards

src/main/java/Task.java
```diff
@@ -0,0 +1,39 @@
...
+        this.isDone = false;
+    }
+
+    public String getStatusIcon() {
+        return (isDone ? "X" : " "); // mark done task with X
+    }
+    public String getTypeIcon() {
+        return "NULL";
+    }
+    public String taskTypeIcon() { return "[" + getTypeIcon() + "]";}
```
jeyvia: Even though these are simple 1 liners, please do still follow the proper style (aka Egyptian Style).

src/main/java/TaskList.java
```diff
@@ -0,0 +1,40 @@
+import java.util.ArrayList;
+
+public class TaskList {
+    private static final ArrayList<Task> TaskList = new ArrayList<>(10);
```
jeyvia, line 6: ArrayList is a type of dynamic array and there is no need to limit your `TaskList` to a size of 10 only!

src/main/java/TaskList.java
```diff
@@ -0,0 +1,40 @@
...
+        return NumTasks;
+    }
+    public static void viewList () {
+        for (int i = 0; i < TaskList.size(); ++i) {
+            System.out.print(i+1 + ". ");
+
+            System.out.println(TaskList.get(i).getTask());
+        }
+    }
+    public static void markDone (int index) {
```
jeyvia, line 39: There should not be a space between `markDone` and `(`. Do change this for your other methods as well :)

src/main/java/Task.java
```diff
@@ -0,0 +1,39 @@
+
+public class Task {
+    protected String description;
+    protected boolean isDone;
+
+    public Task(String description) {   //ok to leave as public?
```
jeyvia: Yes you need to leave it as public. But do remove such comments from your code as it goes against the java coding standard. 👍

PR #205, opened 2023-02-01
[Sebastian Soewanto] iP

SebastianSoewanto:
Up to level 3


src/main/java/Duke.java
```diff
@@ -1,10 +1,53 @@
...
+
 public class Duke {
     public static void main(String[] args) {
         String logo = " ____        _        \n"
                 + "|  _ \\ _   _| | _____ \n"
                 + "| | | | | | | |/ / _ \\\n"
                 + "| |_| | |_| |   <  __/\n"
                 + "|____/ \\__,_|_|\\_\\___|\n";
+
+        String line = "____________________________________________________\n";
```
DavidVin357: Seems like it is a constant, so maybe consider using `static final LINE_DIVIDER` (or some other similar name)?

src/main/java/Duke.java
```diff
@@ -1,10 +1,53 @@
...
         System.out.println("Hello from\n" + logo);
+        System.out.print(line + "Hello! I'm Duke\n" + "What can I do for you?\n" + line);
+        String[] list = new String[100];
```
DavidVin357: I think you use this list to store the different tasks, so maybe it's better to name it as `tasks`? :)

src/main/java/Duke.java
```diff
@@ -1,10 +1,53 @@
...
+        boolean[] isDone = new boolean[100];
+        String toFill;
+        int listCount = 0;
```
DavidVin357: Maybe something like `tasksCount` can tell better on what you are counting here?

src/main/java/Duke.java
```diff
@@ -1,10 +1,53 @@
...
                 + "| | | | | | | |/ / _ \\\n"
                 + "| |_| | |_| |   <  __/\n"
                 + "|____/ \\__,_|_|\\_\\___|\n";
+
+        String line = "____________________________________________________\n";
         System.out.println("Hello from\n" + logo);
+        System.out.print(line + "Hello! I'm Duke\n" + "What can I do for you?\n" + line);
+        String[] list = new String[100];
+        boolean[] isDone = new boolean[100];
+        String toFill;
```
DavidVin357: I think naming it `checkerContent` or some other noun may better represent the meaning of the variable (verbs are more suitable to be used for the methods)

src/main/java/Duke.java
```diff
@@ -1,10 +1,53 @@
...
+            } else if(input.equals("list")) {
+                System.out.print(line + "Here are the tasks in your list:\n");
+                for (int i = 0; i < listCount; i++) {
+                    if (isDone[i]) {
+                        toFill = "X";
+                    }
+                    else {
+                        toFill = " ";
+                    }
+                    System.out.println(i + 1 + ". [" + toFill + "] " + list[i]);
```
DavidVin357: Extracting i+1 to another line as `i++` may make the code a little bit more clean)

DavidVin357 (commented):
LGTM, just several naming proposals :) 
Also, maybe consider `switch case` instead of lots `if else` statements.  Some comments for explaining specific variables might also be nice. 
Overall, it's fine in terms of code standard.

src/main/java/Duke.java
```diff
@@ -8,21 +8,39 @@ public static void main(String[] args) {
                 + "| |_| | |_| |   <  __/\n"
                 + "|____/ \\__,_|_|\\_\\___|\n";
 
-        String line = "_________________________________\n";
+        String line = "____________________________________________________\n";
         System.out.println("Hello from\n" + logo);
         System.out.print(line + "Hello! I'm Duke\n" + "What can I do for you?\n" + line);
         String[] list = new String[100];
```
ghzr0: Naming of  ' String[ ]  list  ' is quite vague, maybe if possible to provide a more useful name for readability? e.g. ' String [ ] taskList '

src/main/java/Duke.java
```diff
@@ -8,21 +8,39 @@ public static void main(String[] args) {
...
                 System.out.print(line + "Bye. Hope to see you again soon!\n" + line);
                 break;
-            } else if(input.equals("list")) {
+            } else if (input.startsWith("mark")) {
+                int mark = Integer.parseInt(input.substring(5));
+                isDone[mark - 1] = true;
+                System.out.println(line + "Nice! I've marked this task as done:\n" + "  [X] " + list[mark - 1]);
+                System.out.print(line);
+            } else if (input.startsWith("unmark")) {
+                int unmark = Integer.parseInt(input.substring(7));
```
ghzr0: The naming of ' unmark' is a bit confusing, maybe switch to  ' taskToUnmark '?

src/main/java/Duke.java
```diff
@@ -8,21 +8,39 @@ public static void main(String[] args) {
...
+                isDone[unmark - 1] = false;
+                System.out.println(line + "OK, I've marked this task as not done yet:\n" + "  [ ] "+ list[unmark - 1]);
                 System.out.print(line);
+            } else if(input.equals("list")) {
```
ghzr0: Good indentation and presenting the conditional statements, which helps the code's readability.

src/main/java/Duke.java
```diff
@@ -8,21 +8,39 @@ public static void main(String[] args) {
                 + "| |_| | |_| |   <  __/\n"
                 + "|____/ \\__,_|_|\\_\\___|\n";
 
-        String line = "_________________________________\n";
+        String line = "____________________________________________________\n";
```
ghzr0: Naming can be changed to avoid confusion down the code block, could change to  -> e.g. 'static final String margin'  (since its a repeating string pattern to be printed out)

ghzr0 (commented):
Commented about code's quality

src/main/java/Duke.java
```diff
@@ -1,10 +1,64 @@
...
+                System.out.print(line);
+            } else if (input.startsWith("todo")) {
+                task[taskCount] = new Todo(input.substring(5));
+                taskCount++;
+                System.out.println(line + "Got it. I've added this task:\n" + task[taskCount - 1]);
+                System.out.print("Now you have " + taskCount + " tasks in the list.\n" + line);
+            } else if (input.startsWith("deadline")) {
+                int byIndex = input.indexOf("/by");
+                task[taskCount] = new Deadline(input.substring(9, byIndex), input.substring(byIndex + 4));
+                taskCount++;
```
bdthanh, line 45: Instead of using taskCount, you can call the size function of the list

src/main/java/Duke.java
```diff
@@ -1,10 +1,64 @@
...
         String logo = " ____        _        \n"
                 + "|  _ \\ _   _| | _____ \n"
                 + "| | | | | | | |/ / _ \\\n"
                 + "| |_| | |_| |   <  __/\n"
                 + "|____/ \\__,_|_|\\_\\___|\n";
+
+        String line = "____________________________________________________\n";
         System.out.println("Hello from\n" + logo);
+        System.out.print(line + "Hello! I'm Duke\n" + "What can I do for you?\n" + line);
+        Task[] task = new Task[100];
```
bdthanh: since the number of task can be more than 100. you can consider to use the dynamic size array list

src/main/java/Duke.java
```diff
@@ -1,10 +1,64 @@
...
+
 public class Duke {
     public static void main(String[] args) {
         String logo = " ____        _        \n"
                 + "|  _ \\ _   _| | _____ \n"
                 + "| | | | | | | |/ / _ \\\n"
                 + "| |_| | |_| |   <  __/\n"
                 + "|____/ \\__,_|_|\\_\\___|\n";
+
+        String line = "____________________________________________________\n";
```
bdthanh: Constant should be named by CAPITAL LETTER for example String DIVIDER, String LOGO

src/main/java/Duke.java
bdthanh: task should be renamed to tasks (with s)

bdthanh (commented):
You should use the correct tag name so that your progress is recorded. Keep up the good work and try your best for all levels. Be careful with naming constants and list variable!
image

PR #206, opened 2023-02-01
[Lyu Xingyu] iP

Zemdalk:


src/main/java/Duke.java
```diff
@@ -1,10 +1,102 @@
...
+        System.out.println(enterGreet);
+        TodoList todoList = new TodoList();
+        while(true){
+            String line = in.nextLine();
+            System.out.println(line);
+            if(line.equals("bye")){
+                // quit
+                System.out.println(exitPrompt);
+                break;
+            }else{
```
lihka1202: Perhaps you can leave a space after the conditional clause and opening brace

src/main/java/Duke.java
```diff
@@ -1,10 +1,102 @@
...
+                    }
+                    todoList.markItem(index, true);
+                }else if(func.equals("unmark")){
+                    // mark work as unfinished
+                    int index;
+                    try {                        
+                        index = Integer.parseInt(line.substring(funcIdx+1, line.length())) - 1;
+                    } catch (Exception e) {
+                        index = -1;
+                    }
```
lihka1202: I like how you have added error handling

src/main/java/Duke.java
```diff
@@ -1,10 +1,102 @@
...
+                // quit
+                System.out.println(exitPrompt);
+                break;
+            }else{
+                // parse input
+                int funcIdx = line.indexOf(" ");
+                if(funcIdx == -1){
+                    funcIdx = line.length();
+                }
+                String func = line.substring(0, funcIdx);
```
lihka1202: perhaps you could give this variable a better name, as it is stored as a `String` but is named `func`

src/main/java/Duke.java
```diff
@@ -1,10 +1,102 @@
...
+                    // mark work as unfinished
+                    int index;
+                    try {                        
+                        index = Integer.parseInt(line.substring(funcIdx+1, line.length())) - 1;
+                    } catch (Exception e) {
+                        index = -1;
+                    }
+                    todoList.markItem(index, false);
+                }else if(func.equals("todo")){
+                    Todo todo = Todo.toTodo(instruction);
```
lihka1202: Perhaps you could name object `todo` better, as it has the same name as the class and it might get confusing!

src/main/java/Duke.java
```diff
@@ -1,10 +1,102 @@
...
+
+                // do instructions
+                if(line.equals("list")){
+                    // show to-do list
+                    todoList.showList();
+                }else if(func.equals("mark")){
+                    // mark work as done
+                    int index;
+                    try {                        
+                        index = Integer.parseInt(line.substring(funcIdx+1, line.length())) - 1;
```
lihka1202: Could you abstract this out to maintain SLAP

src/main/java/Duke.java
TopGun2001: I agree with @lihka1202 . I'd suggest to give a variable name like **commandIndex** or **inputIndex** as it it a bit more clear for readers to understand what you are referring to.

src/main/java/Duke.java
TopGun2001: New implementation I've noticed. I have not seen this before. Maybe something that I should google up. Well done~

src/main/java/Event.java
```diff
@@ -0,0 +1,28 @@
+public class Event extends Task {
+    
+    protected String from;
+    protected String to;
+
+    public static Event toEvent(String instruction){
+        int contentIdx = instruction.indexOf("/from");
+        int fromIdx = instruction.indexOf("/to");
```
TopGun2001: Perhaps u could use **Index** instead of **Idx**

src/main/java/Duke.java
```diff
@@ -51,14 +66,31 @@ public static void main(String[] args) {
...
+                        continue;
+                    }
+                    todoList.addItem(deadline);
+                }else if(func.equals("event")){
+                    Event event = Event.toEvent(instruction);
+                    if(event == null){
+                        printError("Wrong event format");
+                        continue;
+                    }
+                    todoList.addItem(Event.toEvent(instruction));
```
0nandon: 'switch' statement would be better since there are many cases you want to separate.

src/main/java/Deadline.java
```diff
@@ -0,0 +1,23 @@
...
+    public static Deadline toDeadline(String instruction){
+        int contentIdx = instruction.indexOf("/by");
+
+        if (contentIdx == -1) return null;
+        String deadlineContent = instruction.substring(0, contentIdx);
+        String deadlineBy = instruction.substring(contentIdx + "/by ".length());
+        return new Deadline(deadlineContent, deadlineBy);
+    }
+
+    public Deadline(String description, String by) {
```
0nandon:
```suggestion
public getDeadline(String description, String by) {
```

0nandon (commented):
Function is better to be in verb form.

src/main/java/Duke.java
0nandon: There is deep nesting in **if-else** statement while every **if** statements inside are conducting similar calculation. There would be other way to better construct this code like declaring special function for those calculation.

0nandon (commented):
It is better to avoid deep nesting as possible as you can.

src/main/java/TodoList.java
```diff
@@ -12,51 +12,49 @@ public int getListnum(){
...
                 System.out.print("      [X] ");
             }else{
-                System.out.println("    OK, I've marked this task as not done yet:");
+            System.out.println(SPLITTER);
+            System.out.println("    OK, I've marked this task as not done yet:");
                 System.out.print("      [ ] ");
             }
             System.out.println(tasks[num].getDescription());
             System.out.println(SPLITTER);
             System.out.println();
```
0nandon: So many printings, I think in if-else statement. How about just making additional function for those printing stacks?

src/main/java/Duke.java
Zemdalk: Agreed. Thank you for your suggestion!

src/main/java/Duke.java
Zemdalk: Thank you for your good suggestions! I have made corrections.

PR #212, opened 2023-02-02
[Ong Jun Lin Jeremiah] iP

miahjerry:


src/main/java/Duke.java
```diff
@@ -1,10 +1,73 @@
-public class Duke {
+import java.util.Scanner;
+import java.util.*;
```
yixuann02: Perhaps you could list the imported classes explicitly

src/main/java/Duke.java
```diff
@@ -1,10 +1,73 @@
...
+
+public class
+Duke {
```
yixuann02:
```suggestion
public class Duke {
```

src/main/java/Duke.java
```diff
@@ -1,10 +1,73 @@
...
         System.out.println("Hello from\n" + logo);
+
+        String greeting = ("____________________________________________________________\n"
+                + "Hello! I'm Duke\n" + "What can I do for you?\n"
+                + "____________________________________________________________\n"
+        );
+
+        String goodBye = ("____________________________________________________________\n"
+                + "Bye. Hope to see you again soon!\n"
+                + "____________________________________________________________\n");
```
yixuann02: Perhaps you can change your constant names to be all uppercase using underscore to separate words

src/main/java/Duke.java
```diff
@@ -1,10 +1,73 @@
...
+                            + "added:" + userInput + "\n"
+                            + "____________________________________________________________");
+                    toDoList.add(new Task(userInput));
+                }
+            } else {
+                int itemNumber = Integer.parseInt(userInput.replaceAll("[^0-9]", "")) - 1;
+                if (userInput.contains("unmark")) {
+                    toDoList.get(itemNumber).markAsUnDone();
+                    System.out.println("OK, I've marked this task as not done yet:\n" + "    "
+                            + toDoList.get(itemNumber).getStatusIcon() + toDoList.get(itemNumber).getDescription());
```
yixuann02: I think you did a good job indenting wrapped lines.

src/main/java/Duke.java
tyuyang: You may want to refactor the greeting and goodbye into another function instead of putting it all in main(). For example a void printGreeting() function that just prints the greeting

src/main/java/Task.java
```diff
@@ -0,0 +1,29 @@
+public class Task {
+    protected String description;
+    protected boolean isDone;
+    protected int taskNumber;
```
tyuyang: There are no methods that uses taskNumber?

src/main/java/Duke.java
```diff
@@ -1,10 +1,73 @@
...
+        return false;
+    }
+
     public static void main(String[] args) {
         String logo = " ____        _        \n"
                 + "|  _ \\ _   _| | _____ \n"
                 + "| | | | | | | |/ / _ \\\n"
                 + "| |_| | |_| |   <  __/\n"
                 + "|____/ \\__,_|_|\\_\\___|\n";
         System.out.println("Hello from\n" + logo);
```
tyuyang: Can be refactored into print logo function, or you can combine it with the print greeting function

src/main/java/Duke.java
```diff
@@ -1,10 +1,73 @@
...
+                + "____________________________________________________________\n");
+
+        System.out.println(greeting);
+        Scanner myObj = new Scanner(System.in);
+        String userInput;
+        userInput = myObj.nextLine();
+        ArrayList<Task> toDoList = new ArrayList<Task>(100);
+
+        while (!userInput.equals("bye")) {
+            if (!containsNumbers(userInput)) {
```
tyuyang: You probably don't need this if clause. Instead, you can see if the input string contains "list", "mark", or "unmark", and create 3 separate switch cases for them. Even better, you can refactor this entire if else block into a function that parses the user input.

src/main/java/Duke.java
```diff
@@ -1,10 +1,73 @@
...
+                if (userInput.equals("list")) {
+                    System.out.println("Here are the tasks in your list:\n");
+                    for (Task item : toDoList) {
+                        System.out.print((toDoList.indexOf(item) + 1) + ".");
+                        System.out.print(item.getStatusIcon());
+                        System.out.println(item.getDescription());
+                    }
+                    System.out.println("____________________________________________________________\n");
```
tyuyang: You can declare a constant for the row of underscores, or you can just create a function printLine to print the row of underscores. Also, you may want to use System.lineSeparator() instead of '\n' as recommended by the module.

src/main/java/Duke.java
```diff
@@ -1,10 +1,73 @@
...
+        ArrayList<Task> toDoList = new ArrayList<Task>(100);
+
+        while (!userInput.equals("bye")) {
+            if (!containsNumbers(userInput)) {
+                if (userInput.equals("list")) {
+                    System.out.println("Here are the tasks in your list:\n");
+                    for (Task item : toDoList) {
+                        System.out.print((toDoList.indexOf(item) + 1) + ".");
+                        System.out.print(item.getStatusIcon());
+                        System.out.println(item.getDescription());
```
tyuyang: Can be refactored into a separate function printTask(), since all tasks have the same format to be printed. In addition, depending on the situation, you may want to create a new function for each System.out.print() call, if you print the same thing multiple time.

src/main/java/Duke.java
```diff
@@ -1,10 +1,90 @@
-public class Duke {
+import java.util.Scanner;
+import java.util.ArrayList;
+
+public class
+Duke {
```
R-Ramana: Please bring `Duke {` in line 5 to the same line as `public class` (line 4)
```suggestion
public class Duke {
```

src/main/java/Duke.java
```diff
@@ -1,10 +1,90 @@
...
     public static void main(String[] args) {
```
R-Ramana: Your main method is far too long, the recommendation according to the [textbook](https://nus-cs2113-ay2223s2.github.io/website/se-book-adapted/chapters/codeQuality.html#avoid-long-methods) is 30 LOC. Please try to modularize your main method.

PR #213, opened 2023-02-02
[Tze Loong] iP

TzeLoong:


src/main/java/Duke.java
```diff
@@ -1,10 +1,77 @@
+import java.util.Scanner;
+
 public class Duke {
-    public static void main(String[] args) {
```
nicholas132000, line 2: Consistent 4 spaces indent throughout the code, good job

src/main/java/Duke.java
```diff
@@ -1,10 +1,77 @@
...
+
+        scan.close();
+
+    }
+
+    public static void main(String[] args) {
+        greet_user();
+        int counter = 0;
+
+        Tasks[] list_of_tasks = new Tasks[101]; // Following the assumption that there will be no more than 100 tasks
```
nicholas132000: Perhaps you try using camelcase when naming variables. Eg. listOfTasks

src/main/java/Tasks.java
```diff
@@ -0,0 +1,23 @@
+public class Tasks {
```
nicholas132000: Perhaps you could name this class as "Task" (singular) instead as it refers to one task

src/main/java/Duke.java
```diff
@@ -1,10 +1,77 @@
...
                 + "|  _ \\ _   _| | _____ \n"
                 + "| | | | | | | |/ / _ \\\n"
                 + "| |_| | |_| |   <  __/\n"
                 + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
+        System.out.println(logo);
+        System.out.println("How can i help u? \n");
+    }
+
+    public static void echo() {
```
nicholas132000: Good job for using a verb to name this method

src/main/java/Duke.java
```diff
@@ -1,10 +1,77 @@
...
+                System.out
+                        .println(list_of_tasks[index - 1].getStatusIcon() + " " + list_of_tasks[index - 1].description);
+            } else {
+                System.out.println("added: " + input);
+                list_of_tasks[counter].description = input;
+                list_of_tasks[counter].isDone = false;
+                counter++;
+            }
+            scan = new Scanner(System.in);
+            input = scan.nextLine();
```
SongZijin: Since you abstracted the [greet_user()] function, it may be better if you maintain Single Level of Abstraction Principle (SLAP), and also extract this portion as a separate function

src/main/java/Duke.java
SongZijin: It may be better to avoid using a magic number (100) and instead define it at the top of the file as MAX_NUMBER_OF_TASKS

src/main/java/Duke.java
```diff
@@ -1,10 +1,77 @@
...
+            System.out.println(input);
+            echo();
+        }
+
+        scan.close();
+
+    }
+
+    public static void main(String[] args) {
+        greet_user();
```
SongZijin: Good job on extracting this part as a separate function

src/main/java/Duke.java
```diff
@@ -1,10 +1,77 @@
...
+        Scanner scan = new Scanner(System.in);
+        String input = scan.nextLine();
+
+        while (!"bye".equals(input)) {
+            if ("list".equals(input)) {
+                System.out.println("Here are the tasks in your list: ");
+                for (int i = 0; i < counter; i++) {
+                    System.out.println(
+                            i + 1 + "." + " " + list_of_tasks[i].getStatusIcon() + " " + list_of_tasks[i].description);
+                }
```
SongZijin: It may be clearer to extract these parts as separate functions to avoid nesting loops.

SongZijin (commented):
In general, the code is quite clear, however, there is a lot of potential for clearer and more concise code.

src/main/java/Duke.java
```diff
@@ -1,10 +1,106 @@
...
+    public static void main(String[] args) {
+        greet_user();
+
+        Scanner scan = new Scanner(System.in);
+        String input = scan.nextLine();
+        String[] command = input.split(" ", 2);
+
+        while (!"bye".equals(input)) {
+            switch (command[0]) {
+                case "todo":
```
cheehengk: Switch-case should not have different indentations.

src/main/java/Duke.java
```diff
@@ -1,10 +1,106 @@
...
+    }
+
+    public static void List() {
+        System.out.println("Here are all of your " + counter + " tasks: ");
+        for (int i = 0; i < counter; i++) {
+            System.out.println(i + 1 + "." + list_of_tasks[i].toString());
+        }
+    }
+
+    public static void main(String[] args) {
```
cheehengk: Try keeping your main method short and have a parser to deal with interpreting user commands instead

src/main/java/Duke.java
```diff
@@ -1,10 +1,106 @@
...
+                case "todo":
+                    Todo(command[1]);
+                    break;
+                case "deadline":
+                    String[] d = command[1].split("/by", 2);
+                    String d_description = d[0];
+                    String d_by = d[1];
+                    Deadline(d_description, d_by);
+                    break;
+                case "event":
```
cheehengk: Perhaps consider using constants instead of strings, for example, COMMAND_EVENT instead of "event"

src/main/java/Duke.java
```diff
@@ -1,10 +1,106 @@
...
+                    String e_description = e[0];
+                    String e_start = e[1];
+                    String e_end = e[2];
+                    Event(e_description, e_start, e_end);
+                    break;
+                case "list":
+                    List();
+                    break;
+                case "mark":
+                    Integer m_index = Integer.valueOf(command[1]);
```
cheehengk: m_index, u_index etc. might be easy for the original developer to understand but not for someone else reading it or another person taking over. Consider using clearer naming in this case.

src/main/java/Duke.java
```diff
@@ -1,10 +1,106 @@
+import java.util.Scanner;
+
 public class Duke {
-    public static void main(String[] args) {
+
+    private static Tasks[] list_of_tasks = new Tasks[101];
```
cheehengk: Consider other data structures, apart from a fixed size array, to make your product more scalable, especially since we are already approaching recess week :)

src/main/java/Duke.java
```diff
@@ -1,10 +1,106 @@
...
+    }
+
+    public static void Deadline(String description, String by) {
+        Deadline d = new Deadline(description, by);
+        list_of_tasks[counter] = d;
+        counter++;
+        print_action();
+    }
+
+    public static void Event(String description, String start, String end) {
```
cheehengk: Avoid using constructor names of other classes as method name. It can be confusing and potential source of bug. Also, consider separate classes for adding new tasks.

cheehengk (commented):
Overall good effort in implementation.More work on SLAP with shorter methods, and classes will be helpful in future iP levels and tP.

PR #227, opened 2023-02-15
[Lee Zhi Zhong Moses] iP

moseslee9012:


src/main/java/Duke.java
```diff
@@ -1,10 +1,96 @@
...
+import java.util.Arrays;
 public class Duke {
+
+    private static Scanner in = new Scanner(System.in);
+    private static final Task[] tasks = new Task[100];
+
+    public static void command(String line) {
+        String exitCommand = "bye";
+        String toDo = "list";
+        String checkAsDone = "mark";
```
wcwy: You might want to consider converting them into named constants instead of a variables since it is unlikely for them to be changed.
E.g. `public static final String EXIT_COMMAND = "bye";`

src/main/java/Duke.java
```diff
@@ -1,10 +1,96 @@
+import java.util.Scanner;
+import java.util.Arrays;
 public class Duke {
+
+    private static Scanner in = new Scanner(System.in);
+    private static final Task[] tasks = new Task[100];
+
+    public static void command(String line) {
```
wcwy: A better naming practices for methods is to start it with a verb. For example, you can call this `executeCommand` or something better here. (While the word command can be a verb too, it doesn't suit your usage very well)

src/main/java/Duke.java
```diff
@@ -1,10 +1,96 @@
...
+        String exitCommand = "bye";
+        String toDo = "list";
+        String checkAsDone = "mark";
+        int numberOfTasks = 0;
+
+        String[] tokens = line.split(" ", 2);
+        String command = tokens[0];
+
+        while (!command.equals(exitCommand)) {
```
wcwy, line 47: Currently, this entire method is too long. You can try to keep a method short by extracting its content out to other methods whenever appropriate.

src/main/java/Duke.java
```diff
@@ -1,10 +1,96 @@
...
+                tokens = line.split(" ", 2);
+                command = tokens[0];
+            } else if (command.equals(checkAsDone)){
+                int finishedTask = Integer.parseInt(tokens[1]) - 1;
+                tasks[finishedTask].markAsDone();
+                System.out.println("Alright! I have marked the task as complete!");
+
+                for (int i = 0; i < numberOfTasks; ++i) {
+                    System.out.format("%d. ", (i + 1));
+                    System.out.println(tasks[i]);
```
wcwy, line 64: This is a good use of polymorphism here. You can apply a similar concept when dealing with similar scenarios in further levels.

src/main/java/ToDo.java
```diff
@@ -0,0 +1,11 @@
+public class ToDo extends Task {
+
+    public ToDo(String description) {
+        super(description);
+    }
+
+    @Override
+    public String toString() {
+        return "[T]" + super.toString();
```
wcwy, line 9: Arguably, strings like `[T]` can be seen as a magic strings. You might want to consider extracting strings like this into a named constants. Sometimes, you could also use enums to achieve similar effect. There could be other magic strings/numbers/literals in your other code too that you might want to look out for them.

src/main/java/Duke.java
```diff
@@ -1,10 +1,96 @@
...
+                }
+
+                line = in.nextLine();
+                tokens = line.split(" ", 0);
+                command = tokens[0];
+            } else {
+                String task = tokens[1];
+                switch(command) {
+                case "todo":
+                    Task t = new ToDo(task);
```
wcwy: We want to avoid using a single character as the name of a variable (with some exceptions such as i, j, k in a for loop). Try to think if another name for this

src/main/java/Duke.java
```diff
@@ -1,10 +1,96 @@
+import java.util.Scanner;
+import java.util.Arrays;
 public class Duke {
+
+    private static Scanner in = new Scanner(System.in);
+    private static final Task[] tasks = new Task[100];
```
wcwy, line 6: For program scalability, you might want to consider using `ArrayList` instead of setting a maximum number of tasks user can have here.

src/main/java/Task.java
```diff
@@ -0,0 +1,21 @@
...
+    protected String description;
+    protected boolean isDone;
+
+    public Task(String description) {
+        this.description = description;
+        this.isDone = false;
+    }
+
+    public String getStatus() {
+        return (isDone ? "[X]" : "[ ]"); // mark done task with X
```
wcwy, line 11: You might want to capitalise the first character in your comment.

wcwy (commented):
Well done in catching up in the past one week! Keep the spirit up! So far most of your code follows well in terms of coding standards and coding quality. You might want to consider extracting methods out more often whenever appropriate to prevent one single method to be too long and to avoid issues such as arrow code. Below are some suggestions on the possible things you can look out for. All the best for further levels!

PR #230, opened 2023-02-17
[Bingyuan] iP

notbingsu:
Week 4 PR


src/main/java/Task.java
```diff
@@ -0,0 +1,36 @@
...
+    protected boolean isDone;
+
+    public Task(String description) {
+        this.description = description;
+        this.isDone = false;
+    }
+
+    public String getStatusIcon() {
+        return (isDone ? "X" : " "); // mark done task with X
+    }
```
theopin, line 16: Do consider replacing such string literals with named constants, it will help in enhancing readability of your code

src/main/java/Main.java
```diff
@@ -0,0 +1,47 @@
...
+                    Duke.addEvent(phrase);
+                } else if (command.equals("todo")) {
+                    Duke.addTodo(phrase);
+                } else if (command.equals("deadline")) {
+                    Duke.addDeadline(phrase);
+                } else if (command.equals("delete")) {
+                    Duke.delete(Integer.parseInt(phrase));
+                } else {
+                    System.out.println("Invalid command");
+                }
```
theopin: Main seems to be overloaded with a lot of functionality, perhaps consider abstracting behaviour such as parsing command (like this block) into another method?

src/main/java/Duke.java
```diff
@@ -1,10 +1,155 @@
...
+        System.out.println("These are the tasks from your file:\n");
+        list();
+    }
+
+    private void addFileData(String[] inputArgs) {
+        Task newTask;
+        String command = inputArgs[0];
+        boolean taskStatus = Boolean.parseBoolean(inputArgs[1]);
+        switch (command) {
+            case "T":
```
theopin: case line should not be tabbed (ie. on the same level as switch)

src/main/java/Duke.java
```diff
@@ -1,10 +1,155 @@
...
+        } catch (IOException e) {
+            System.out.print("\nError getting file data");
+        }
+        System.out.println("These are the tasks from your file:\n");
+        list();
+    }
+
+    private void addFileData(String[] inputArgs) {
+        Task newTask;
+        String command = inputArgs[0];
```
theopin: Do consider renaming inputArgs to its full form for better readability (ie. inputArguments)

src/main/java/Duke.java
```diff
@@ -1,10 +1,155 @@
...
+                    addFileData(inputArgs);
+                }
+                fileData.close();
+            }
+        } catch (IOException e) {
+            System.out.print("\nError getting file data");
+        }
+        System.out.println("These are the tasks from your file:\n");
+        list();
+    }
```
theopin: try-catch should have a finally clause defined to mark the relevant section of code (in this case 117-118) that runs regardless of the outcome of try catch

theopin (commented):
Good job on implementation of duke so far! Some general pointers to improve your code:
- magic literals. Avoid passing in string/int literals into functions directly as it can be difficult for
readers to understand your code. Do consider refactoring them into named constants
- Main method seems to be too long. Perhaps consider trying to break it into segments and then refactoring them into seperate methods in order to better visualize its overall flow.

